Issue #12, opened 2021-03-08
version pattern should match pattern used by check-repo-update


sbouchex:
If the plug in version is 2021.03.0, it should rejected as this version will be rejected afterwards by check-repo-update (when creating a PR)

The regexp should be something like :` ^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*$|\\%VERSION\\%)` to validate the version

marcusolsson:
Thanks! Could you open the pull request and change the schema in the grafana repo (which this tool downloads)?

https://github.com/grafana/grafana/blob/master/docs/sources/developers/plugins/plugin.schema.json

sbouchex:
@marcusolsson PR done

PR #13, opened 2021-03-08
#12 Modified regexp

sbouchex:
Modified regexp to follow npm version semantic


marcusolsson:
Thanks for this! Unfortunately, the schema in this repository is not used anymore (we should remove it to avoid confusion). The plugin validator downloads the schema at https://github.com/grafana/grafana/blob/master/docs/sources/developers/plugins/plugin.schema.json

Could you open your pull request against the grafana/grafana repository instead?

PR #18, opened 2021-03-29
Version 2 of the validator

marcusolsson:
This PR refactors the validator into a design as inspired by the go/analysis package. This has a few benefits:

- Analyzers can depend on other analyzers
- Analyzers can use the result of a previous analyzer
- Each analyzer is extracted into its own package for better isolation

For now, this lives in parallel to the first version to allow it to be tested, and to be able to update the validator UI separately.


briangann (approved):
this looks great, will test it in the new pipeline

PR #19, opened 2021-04-07
Add flag to report private signatures check error as warning

mbialon:
I would like to validate a private plugin so I've added a flag to report private signature check error as warning.

Is this an acceptable solution?


marcusolsson (approved):
LGTM. 

For plugincheck2 we're looking to make this much more configurable so that you can enable/disable as well as change the severity for any check.

PR #20, opened 2021-04-09
Make checks configurable for plugincheck2

marcusolsson:
plugincheck2 now accepts a yaml config to enable/disable analyzers and rules within them.

The major additions are:

- `pass.Reportf` takes a reference to the rule it reports and only reports the diagnostic if it's been enabled
- `initAnalyzers` configures the rules by updating the global variables in each analyzer package. Not proud of this one (a little ashamed in fact). 

Also, we should finalize the names of analyzers and rules to avoid breaking people's configs.


briangann (approved):
lgtm, will try it out

PR #24, opened 2021-07-15
Add docker image

cawolf:
Hi,

after releasing a plugin, I use this tool to validate the distribution zip archive before submitting the new version to the grafana index. My plugin development workflow now looks like this:

* change code, push to feature branch, code review, merge to main
* Github Actions:
  * building the plugin
  * signing the plugin
  * creating the zip archive
  * releasing a new version

What I would like to have is a way to validate the plugin as part of the workflow, in order to not release a version that will never make it to the grafana plugin index:

* change code, push to feature branch, code review, merge to main
* Github Actions:
  * building the plugin
  * signing the plugin
  * creating the zip archive
  * **validating the zip archive**
  * releasing a new version

Now, I could do this in a Github Action directly by installing the plugin-validator in a job. Instead, I would like to have a docker image to serve as a base for a dedicated Github Action, which does the validation part for me. (Depending on the variations of plugin building, the Github Action could possibly extended to handle the full cycle of plugin release, but this is currently out of scope).

---

This PR includes:
* a Dockerfile for the image
* Github Actions that automatically build an image and push it to the Github registry ([see in my fork](https://github.com/cawolf/plugin-validator/pkgs/container/plugin-validator)) whenever a commit is tagged (usually during release)
* documentation on how to use the image (keep in mind: when testing it with my fork, replace `grafana` with `cawolf` in the image id)

I'm looking forward to receiving your feedback. Feel free to question the approach itself, or give me feedback on implementation details. Thanks for your time,
best regards,
@cawolf


jeschkies:
I very much would like to see a GitHub action that runs the plugin. Maybe that should be a follow up.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
jeschkies, line 16: Why is this required?

jeschkies (commented):
:wave: I've introduced a GitHub action in the meanwhile. However, it' still great to push the image. Please see my comment below.

.github/workflows/build.yaml
jeschkies: Ah. Is it a requirement to build for arm etc?

cawolf: Yes, in combination with buildx you can build images for arm as well.

cawolf:
Hi, do you need anything else changed in order to merge this PR?

briangann:
sorry for the delay, i'm taking a look at the PR - this is a great addition

jeschkies:
@briangann shall we merge this change?

academo:
We will provide updates to the existing docker image https://hub.docker.com/r/grafana/plugin-validator-cli

PR #27, opened 2021-11-30
fixes invalid zip content

briangann:
* update to go 1.17
* update packages
* handle plugin id and version detection
* handle malformed zip files
* add testdata


pkg/cmd/plugincheck2/main.go
```diff
@@ -15,6 +15,12 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type FormattedOutput struct {
+	ID         string                           `json:"id"`
+	Version    string                           `json:"version"`
+	Diagnotics map[string][]analysis.Diagnostic `json:"plugin-validator"`
```
marcusolsson, line 21:
```suggestion
	Diagnostics map[string][]analysis.Diagnostic `json:"plugin-validator"`
```

pkg/cmd/plugincheck2/quickmeta.go
```diff
@@ -0,0 +1,30 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// GetIDAndVersion Gets plugin id and version from extracted zip
+func GetIDAndVersion(archiveDir string) (*string, *string, error) {
+	fis, _ := ioutil.ReadDir(archiveDir)
```
marcusolsson, line 13: Are we certain the archive directory exists here? If it doesn't, it'll panic on the next line.

pkg/cmd/plugincheck2/quickmeta.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// GetIDAndVersion Gets plugin id and version from extracted zip
+func GetIDAndVersion(archiveDir string) (*string, *string, error) {
```
marcusolsson, line 12: Why are we returning references to strings instead of values?

briangann: doesn't have to, was just returning nil/nil/error vs "","",err

pkg/cmd/plugincheck2/quickmeta.go
briangann: good catch will update

briangann: redid this section to handle more edge cases

pkg/cmd/plugincheck2/quickmeta.go
briangann: updated this also

marcusolsson (approved):

PR #32, opened 2022-04-27
Validate that logos and screenshots exists

marefr:
Seems to work as expected, but not sure if some of the changes I've made make sense. Please have a detailed look at the code changes :smile: 

```
❯ ./bin/linux_amd64/plugincheck2 -strict -config config/strict.yaml ~/Downloads/orchestracities-map-panel-1.4.3.zip
error: plugin.json: screenshot path doesn't exists: example.png

```

### DIFF
--- a/pkg/analysis/passes/logos/logos.go
+++ b/pkg/analysis/passes/logos/logos.go
@@ -2,15 +2,21 @@ package logos
 
 import (
 	"encoding/json"
+	"strings"
 
 	"github.com/grafana/plugin-validator/pkg/analysis"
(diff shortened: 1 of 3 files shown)


marcusolsson (approved):
Looks good! No need to update plugincheck imo.

briangann:
you can bump this version also and i can push up a new image https://github.com/grafana/plugin-validator/blob/master/Magefile.go#L22

marefr:
@briangann to "v2"? How's that version related to https://github.com/grafana/plugin-validator/tags?

PR #33, opened 2022-05-03
Pass package path in GitHub action.

jeschkies:


action.yml
```diff
@@ -5,9 +5,14 @@ inputs:
     description: 'Path to configuration file'
     required: false
     default: '/app/config/default.yaml'
+  package:
+    description: 'Path to plugin package file'
+    required: false
+    default: 'plugin.zip'
```
sd2k, line 11: I'm not sure we can have a default since the plugin zip needs to be named according to its ID - pretty sure this is going to be required 😞 

```suggestion
    required: true
```

PR #40, opened 2022-11-29
Expand error messages to include a message detail

academo:
* Modifies the reporting system to include an error "detail" that analyzers can pass.
   This modification required to modify some analyzers that were using the extra parameters in report as a way to do a formatted string
* Expands the error messages for all passes to include a detail of the error.

Fixes https://github.com/grafana/plugin-validator/issues/39


pkg/analysis/passes/jargon/jargon_test.go
```diff
@@ -0,0 +1,65 @@
...
+		Report: func(_ string, d analysis.Diagnostic) {
+			assert.Equal(t, "README.md contains developer jargon: (yarn)", d.Title)
+		},
+	}
+
+	_, err := Analyzer.Run(pass)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
```
xnyo, line 65: We also need to make sure that the `Report` anonymous function runs (which also contains the assertion), otherwise the test will pass if `Report` is not called (and the assertion about "contains yarn" won't be evaluated either):

```suggestion
func TestWithJargon(t *testing.T) {
	var invoked bool

	pass := &analysis.Pass{
		ResultOf: map[*analysis.Analyzer]interface{}{
			readme.Analyzer: jargonReadme,
			Analyzer:        nil,
		},
		Report: func(_ string, d analysis.Diagnostic) {
			invoked = true
			assert.Equal(t, "README.md contains developer jargon: (yarn)", d.Title)
		},
	}

	_, err := Analyzer.Run(pass)
	if err != nil {
		t.Fatal(err)
	}
	assert.True(t, invoked, "report should be called")
}
```

Considering this quite a common thing to do (checking if it has been invoked and running the corresponding function, there's also the same in `TestClean`), you can consider adding some helpers:

```go
// pkg/testcounter/testcounter.go
package testcounter

// FuncCallCounter is a type that implements a counter that can be used to count function invocations in tests.
// The counter can be incremented or reset to zero, but it can't be set to arbitrary values.
type FuncCallCounter struct {
	count int
}

// Inc increments the value of the counter by 1.
func (r *FuncCallCounter) Inc() {
	r.count++
}

// Value returns the current value of the counter.
func (r *FuncCallCounter) Value() int {
	return r.count
}

// Called returns true if the counter was incremented at least once.
func (r *FuncCallCounter) Called() bool {
	return r.count > 0
}

// CalledOnce returns true if the counter was incremented exactly once.
func (r *FuncCallCounter) CalledOnce() bool {
	return r.count == 1
}

// Reset resets the counter to zero.
func (r *FuncCallCounter) Reset() {
	r.count = 0
}
```

```go
// pkg/analysis/testinterceptor.go
package analysis

import "github.com/grafana/plugin-validator/pkg/testcounter"

// ReportCounter allows to create functions to be passed as "report" (func(string, Diagnostic)),
// that will also allow to count how many times the function was called.
type ReportCounter struct {
	testcounter.FuncCallCounter
}

// Function takes a function for a Report in a Pass, calls it and then increments the internal FuncCallCounter.
func (i *ReportCounter) Function(f func(string, Diagnostic)) func(string, Diagnostic) {
	return func(s string, d Diagnostic) {
		f(s, d)
		i.Inc()
	}
}
```

Then the `TestWithJargon` test becomes:

```go
func TestWithJargon(t *testing.T) {
	// This "intercept" variable allows us to track how many times a func(s string, d analysis.Diagnostic) is called.
	var intercept analysis.ReportCounter

	pass := &analysis.Pass{
		ResultOf: map[*analysis.Analyzer]interface{}{
			readme.Analyzer: jargonReadme,
			Analyzer:        nil,
		},
		// The func returned by "intercept.Function()" will call the passed func (the one with the "assert")
		// and also increment its internal counter, allowing us to assert on how many times the func was called.
		Report: intercept.Function(func(s string, d analysis.Diagnostic) {
			assert.Equal(t, "README.md contains developer jargon: (yarn)", d.Title)
		}),
	}

	_, err := Analyzer.Run(pass)
	if err != nil {
		t.Fatal(err)
	}
	
	// Assert on how many times the function was called.
	assert.True(t, intercept.CalledOnce(), "report should be called once")
}
```

Maybe a bit too overkill? If you have to do this a lot, I think it's worth checking it out (for a follow-up PR, if you plan to write tests for everything). It's also a good exercise on anonymous functions in go if you can figure out what's going on :)

pkg/analysis/passes/legacyplatform/legacyplatform.go
```diff
@@ -27,11 +27,11 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	)
 
 	if angularExp.Match(module) && !reactExp.Match(module) {
-		pass.Reportf(pass.AnalyzerName, legacyPlatform, "module.js: uses legacy plugin platform")
+		pass.ReportResult(pass.AnalyzerName, legacyPlatform, "module.js: uses legacy plugin platform", "The plugin uses the legacy plugin platform (angularjs). Please migrate the plugin to use the new plugins platform.")
 	} else {
 		if legacyPlatform.ReportAll {
 			legacyPlatform.Severity = analysis.OK
-			pass.Reportf(pass.AnalyzerName, legacyPlatform, "module.js: uses current plugin platform")
+			pass.ReportResult(pass.AnalyzerName, legacyPlatform, "module.js: uses current plugin platform", "")
```
xnyo, line 34: (Not related to the new messages) I see this pattern repeated a lot across the various passes we have:

```go
x.Severity = analysis.OK
pass.ReportResult(pass.AnalyzerName, x, "some message", "")
```

(setting severity to OK and empty string as details).

I feel like we can combine those in a single function call? Maybe in another PR? I'll see if we can come up with something

pkg/analysis/analysis.go
```diff
@@ -1,8 +1,6 @@
 package analysis
 
-import (
-	"fmt"
-)
+import ()
```
xnyo, line 4: This line is not necessary :)

```suggestion

```

xnyo (changes requested):
Great job!! 👏👏👏👏
I have added some comments and some stuff for future PRs (not necessarily related to the error message changes, but that can improve the codebase)

Magefile.go
```diff
@@ -185,3 +185,17 @@ func (Run) V2() error {
 		"https://github.com/marcusolsson/grafana-jsonapi-datasource/releases/download/v0.6.0/marcusolsson-json-datasource-0.6.0.zip",
 	)
 }
+
+func (Run) V2Local(ctx context.Context, path string) error {
+	mg.Deps(Build.Local)
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
academo: I'll leave it as ```	if _, err := os.Stat(path); err != nil {
		return err
	}
``` We don't need to handle that many error cases.

pkg/analysis/passes/jargon/jargon_test.go
academo: I appreciate the suggestion for the helpers. I'll take them in consideration for the next PR where I'll improve on the tests for all the passes. So far I pushed the check for invoked.

pkg/analysis/passes/legacyplatform/legacyplatform.go
academo: Indeed. We could have "ReportIssue" and "ReportOK" functions.

pkg/analysis/passes/jargon/jargon_test.go
xnyo: :+1: Agree on keeping it in a separate PR if necessary

pkg/analysis/passes/legacyplatform/legacyplatform.go
xnyo: :+1: Agree

xnyo (approved):
LGTM! :rocket: 
The other suggestions can be discussed/addressed in a separate PR as they are not strictly related to error messages

PR #43, opened 2022-12-13
Add tests for modulejs, legacyplatform, logos and manifest passes

academo:
Tests for:

* modulejs
* legacyplatform
* logos
* manifest


briangann (approved):
LGTM!

PR #45, opened 2022-12-21
Adjust error severity types on each validator

academo:
- Use a default severity to warning if none is passed
- Allow rules to define their own severity and use them if not overwritten by config
- Adding default severities for all passes

Note to reviewer: Even though they seem to be many file changes they are repetitive.


xnyo (approved):
LGTM! :sparkles:

PR #48, opened 2023-01-04
Add validation for executable existance and permissions (if defined)

academo:
Closes https://github.com/grafana/grafana/issues/52640

Adds a validator for the archive binary files if there's a defined `executable` in the plugin.json file

### DIFF
--- a/pkg/analysis/passes/binarypermissions/binarypermissions.go
+++ b/pkg/analysis/passes/binarypermissions/binarypermissions.go
@@ -23,7 +23,7 @@ var Analyzer = &analysis.Analyzer{
 	Name:     "archivename",
 	Requires: []*analysis.Analyzer{metadata.Analyzer, archive.Analyzer},
 	Run:      run,
-	Rules:    []*analysis.Rule{binaryExecutableFound},
+	Rules:    []*analysis.Rule{binaryExecutableFound, binaryExecutablePermissions},
 }
 
 var binarySuffixes = []string{
@@ -43,7 +43,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		return nil, err
 	}
 
-	if data.Executable == "" {
+	if data.Executable == "" && binaryExecutableFound.ReportAll {
 		binaryExecutableFound.Severity = analysis.OK
 		pass.ReportResult(pass.AnalyzerName, binaryExecutableFound, "No executable defined in plugin.json", "")
 		return nil, nil
@@ -81,16 +81,16 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		if err != nil {
 			pass.ReportResult(pass.AnalyzerName,
 				binaryExecutablePermissions,
-				fmt.Sprintf("Could not read permissions for executable %s", executable),
-				fmt.Sprintf("Could not read the file %s. This could be too restrictive in your binary files or your zip file is corrupted", executable))
+				fmt.Sprintf("Could not read permissions for executable %s", binaryPath),
+				fmt.Sprintf("Could not read the file %s. This could be too restrictive in your binary files or your zip file is corrupted", binaryPath))
 		}
 
 		filePermissions := fileInfo.Mode().Perm()
 		if filePermissions != REQUIRED_PERMISSIONS {
 			pass.ReportResult(pass.AnalyzerName,
 				binaryExecutablePermissions,
-				fmt.Sprintf("Permissions for binary executable %s are incorrect (%04o found).", executable, filePermissions),
-				fmt.Sprintf("The binary file %s must have exact permissions %04o (%s).", executable, REQUIRED_PERMISSIONS, REQUIRED_PERMISSIONS))
+				fmt.Sprintf("Permissions for binary executable %s are incorrect (%04o found).", filepath.Base(binaryPath), filePermissions),
+				fmt.Sprintf("The binary file %s must have exact permissions %04o (%s).", filepath.Base(binaryPath), REQUIRED_PERMISSIONS, REQUIRED_PERMISSIONS))
 		}
 	}
 
--- a/pkg/analysis/passes/binarypermissions/binarypermissions.go
+++ b/pkg/analysis/passes/binarypermissions/binarypermissions.go
@@ -1,0 +1,98 @@
+package binarypermissions
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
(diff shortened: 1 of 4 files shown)


pkg/analysis/passes/metadata/types.go
```diff
@@ -1,10 +1,11 @@
...
 type Metadata struct {
-	ID   string       `json:"id"`
-	Name string       `json:"name"`
-	Type string       `json:"type"`
-	Info MetadataInfo `json:"info"`
+	ID         string       `json:"id"`
+	Name       string       `json:"name"`
+	Type       string       `json:"type"`
+	Info       MetadataInfo `json:"info"`
+	Executable string       `json:"executable"`
```
academo, line 8: `Executable` added here.

tolzhabayev (approved):
Awesome!

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/archive"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/metadata"
+)
+
+var (
+	binaryExecutableFound       = &analysis.Rule{Name: "binary-executable-found", Severity: analysis.Error}
+	binaryExecutablePermissions = &analysis.Rule{Name: "binary-executable-permissions", Severity: analysis.Error}
+)
+
+var REQUIRED_PERMISSIONS = fs.FileMode.Perm(0755)
```
xnyo, line 20: The Go convention does not use snake case, use camel case is used instead, even if it acts like a constant. This should become:

```go
var requiredPermissions = fs.FileMode.Perm(0755)
```

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -0,0 +1,99 @@
...
+	var data metadata.Metadata
+	if err := json.Unmarshal(metadataBody, &data); err != nil {
+		return nil, err
+	}
+
+	if data.Executable == "" && binaryExecutableFound.ReportAll {
+		binaryExecutableFound.Severity = analysis.OK
+		pass.ReportResult(pass.AnalyzerName, binaryExecutableFound, "No executable defined in plugin.json", "")
+		return nil, nil
+	}
```
xnyo, line 50: I am not too sure what `ReportAll` is, but by looking at other usages it looks like it's a flag that determines if OK passes are reported or not.

In that case, if `ReportAll` is false and there's no executable, the checks for the executable are still performed. If so, I think this should become:

```suggestion
	if data.Executable == "" {
		if binaryExecutableFound.ReportAll {
			binaryExecutableFound.Severity = analysis.OK
			pass.ReportResult(pass.AnalyzerName, binaryExecutableFound, "No executable defined in plugin.json", "")
		}
		return nil, nil
	}
```

(`return` is required only if `data.Executable == ""`)

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -0,0 +1,99 @@
...
+
+	executable := data.Executable
+	executableParentDir := filepath.Join(archiveDir, filepath.Dir(executable))
+	executableName := filepath.Base(executable)
+
+	var foundBinaries = []string{}
```
xnyo, line 56: Nit/tip: this can become

```go
var foundBinaries []string
```

The zero-value for a slice is nil, but in this case it behaves in the same way.

Just personal preference, both are fine :)

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, suffix := range binarySuffixes {
+		binaryPath := filepath.Join(executableParentDir, executableName+suffix)
+		if _, err := os.Stat(binaryPath); err == nil {
```
xnyo, line 59: Shouldn't this be something like:

```suggestion
		if _, err := os.Stat(binaryPath); err != nil {
			if os.IsNotExist(err) {
				continue
			}
			return nil, fmt.Errorf("stat %q: %w", binaryPath, err)
		}
```

?

Also, unless I flip `err == nil` to `err != nil` all the tests in `package binarypermissions` are failing for me, so I think you made a typo there :)

pkg/analysis/passes/binarypermissions/binarypermissions_test.go
```diff
@@ -0,0 +1,123 @@
...
+	var interceptor testpassinterceptor.TestPassInterceptor
+	pluginJsonContent := []byte(`{
+		"id": "` + pluginId + `",
+		"type": "panel",
+		"executable": "test-plugin-panel"
+	}`)
+
+	// due to CI running on linux, we need to re-create the correct permissions
+	testDataContainer := filepath.Join("testdata", "correct-permissions")
+	os.Chmod(filepath.Join(testDataContainer, "test-plugin-panel_linux_amd64"), 0755)
```
xnyo, line 27: `os.Chmod` returns an `error`. I would handle it just to be safe, like this:

```suggestion
	require.NoError(t, os.Chmod(filepath.Join(testDataContainer, "test-plugin-panel_linux_amd64"), 0755))
```

Same for the other occurrences

xnyo (changes requested):
Amazing! :confetti_ball: 

Tests are failing for me, but I think it's due to a typo in your code (see the comments).

I have also left some tips/nit for free, as usual :smile:

pkg/analysis/passes/binarypermissions/binarypermissions.go
academo: Good catch!

pkg/analysis/passes/binarypermissions/binarypermissions.go
academo: A typo on my last commit update. thanks for spotting. I'll keep the continue even on non `IsNotExist` error.

pkg/analysis/passes/binarypermissions/binarypermissions_test.go
academo: That's a nice addition to debug a broken test in case the OS has a different permission system. I'll add it.

pkg/analysis/passes/binarypermissions/binarypermissions.go
academo: I like to write explicit code rather than implicit. :)

xnyo (approved):
LGTM! :clap: :clap: :clap: :clap:

Issue #35, opened 2022-09-05
Deprecate plugincheck in favor of plugincheck2


tolzhabayev:
- Remove `plugincheck` from the readme and only keep `plugincheck2`
- Show a message to migrate to `plugincheck2` every time it's ran. Do not remove code yet.
- Add a deprecation notice to the `readme.md` of the `plugincheck` pkg
- Mark code as deprecated https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go
- Search for usages of `plugincheck` in the wild:
- - https://cs.github.com/?scopeName=All+repos&scope=&q=plugincheck+language%3AYAML#
- - At least 14 plugins are currently using it

academo:
@tolzhabayev since both of them are compatible can we start putting the plugincheck2 binary in place of plugincheck after a while?

tolzhabayev:
Are they really replaceable / a drop-in replacement?

academo:
> Are they really replaceable / a drop-in replacement?

The parameters are the same and the share the same configuration. I am not really sure why they were made as different projects. I'll ask around.

tolzhabayev:
Ideally we should end up with one project that is versioned (like our go sdk) - we need to figure out what is better:
- Keep `plugincheck` and release a new version of it and deprecate `plugincheck2`
     - Means we would have to migrate everyone and their CI's, etc. to `plugincheck`
- Keep  `plugincheck2` and deprecate `plugincheck` which I would not change at all, also not place another binary into it - just leave as is
     - Means we would have to keep `plugincheck2` name forever and iterate on it using versioning. Naming might be a bit confusing.
     - Migrate everyone to use `plugincheck2`

PR #50, opened 2023-01-05
Deprecate plugincheck v1

academo:
Closes https://github.com/grafana/plugin-validator/issues/35

- Remove plugincheck v1 from readme
- Add ascii deprecated message


pkg/cmd/plugincheck/main.go
```diff
@@ -13,7 +13,29 @@ import (
...
+ .d88888 .d88b. 88888b. 888d888 .d88b.  .d8888b 8888b. 888888 .d88b.  .d88888
+d88" 888d8P  Y8b888 "88b888P"  d8P  Y8bd88P"       "88b888   d8P  Y8bd88" 888
+888  88888888888888  888888    88888888888     .d888888888   88888888888  888
+Y88b 888Y8b.    888 d88P888    Y8b.    Y88b.   888  888Y88b. Y8b.    Y88b 888
+ "Y88888 "Y8888 88888P" 888     "Y8888  "Y8888P"Y888888 "Y888 "Y8888  "Y88888
+                888
+                888
+                888
+
+	**plugincheck V1 is deprecated and you should not use it.
```
leventebalogh, line 33: Nit: do we need a closing `**` as well to make it bold? (In case that was the intention)

leventebalogh (approved):
Nice! 👏 

Probably a dumb question, but all of our pipelines are using the V2 now, right?

pkg/cmd/plugincheck/main.go
academo: Not really this is not supposed to be markdown, just a way to catch attention in the terminal

academo:
> Nice! clap
> 
> Probably a dumb question, but all of our pipelines are using the V2 now, right?

The community plugin pipeline is using v2. I'll have to check if we have other plugins around using this in their github action and using v1.

Issue #51, opened 2023-01-09
Validation improvement: Validate the plugin.json version is higher than the current publish version if the plugin is published


academo:
**What to do if version >= published version?**
Return an error stating that you can not submit an update that has the same or lower version than the one that is published in our catalog.

PR #53, opened 2023-01-10
Add validation for plugin version 

academo:
Closes https://github.com/grafana/plugin-validator/issues/51

Validates that a checked plugin version is higher than the current published version

Example checking clock-panel current zip against the avlidator

image


xnyo (changes requested):
Awesome work :clap::clap::clap::clap:

Left some comments & Go tips :smiley:

pkg/analysis/passes/version/version.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		// in case of any error getting the online status, skip this check
+		return nil, nil
+	}
+	grafanaComVersion := pluginStatus.Version
+	parsedGrafanaVersion, err := version.NewVersion(grafanaComVersion)
+	if err != nil {
+		// in case of any error parsing the online status, skip this check
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
academo: we can't really log these in a meaningful way. If a dev runs this locally they will see an error they can't act on and they don't care. If this runs in CI we don't have anywhere to report it in a meaningful way.

pkg/analysis/passes/version/version.go
```diff
@@ -98,18 +102,12 @@ func getPluginDataFromGrafanaCom(pluginId string) (*PluginStatus, error) {
...
-	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("wrong status code, expected 200 got %d", response.StatusCode)
 	}
 
 	status := PluginStatus{}
-
-	err = json.Unmarshal(content, &status)
-	if err != nil {
+	defer response.Body.Close()
```
xnyo, line 109: My bad, sorry this `Close` should go above, right after the `if err != nil` block on line 94 (so on line 97). This is because we are returning early if `response.StatusCode != http.StatusOK` or `response.StatusCode == http.StatusNotFound`, and in that case we would not close the response body

xnyo (changes requested):
One small thing left:

pkg/analysis/passes/version/version.go
academo: makes sense!

xnyo (approved):
LGTM! :clap::clap::clap:

PR #57, opened 2023-01-18
Add OSV scanner

### LINKED ISSUE #38: Plug-in OSV-scanner for source code dependencies check

briangann:
This adds the osv-scanner as an additional check, it will detect go/yarn lockfiles and scan for vulnerabilities.

Only critical vulnerabilities will stop a build from being published (it will fail validation)

- [x] Magefile needs to be updated to build.local as intended (GOOS determined)
- [x] Update docs on using this feature
- [x] Needs to handle items with no aliases (all critical will have an alias)
- [x] Needs path/arg/env to source code (should not run against a zip file), repo or local
- [x] Needs to clone repo (if not local) and clean up
- [x] Add filtering for well-known packages being pulled in


pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+	// we're detecting only go.mod and package.lock (this can be changed to use defaults)
+	lockFile := filepath.Join(archiveDir, "go.mod")
+	if _, err := os.Stat(lockFile); err != nil {
+		// check for yarn.lock
+		lockFile = filepath.Join(archiveDir, "yarn.lock")
+		if _, err := os.Stat(lockFile); err != nil {
+			// nothing to do... skip
+			scanningSucceeded.Severity = analysis.OK
+			if scanningSucceeded.ReportAll {
+				pass.ReportResult(pass.AnalyzerName, scanningSucceeded, "osv-scannner skipped", "There were no go.mod or yarn.lock files to check")
```
academo, line 52: missing the `yarn.lock` check.

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	archiveDir := pass.ResultOf[archive.Analyzer].(string)
+	// we're detecting only go.mod and package.lock (this can be changed to use defaults)
+	lockFile := filepath.Join(archiveDir, "go.mod")
+	if _, err := os.Stat(lockFile); err != nil {
+		// check for yarn.lock
+		lockFile = filepath.Join(archiveDir, "yarn.lock")
+		if _, err := os.Stat(lockFile); err != nil {
```
academo, line 48: should this check return here since there's nothing the validator can do if there's no go.mod or yarn.lock to check?

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+			// nothing to do... skip
+			scanningSucceeded.Severity = analysis.OK
+			if scanningSucceeded.ReportAll {
+				pass.ReportResult(pass.AnalyzerName, scanningSucceeded, "osv-scannner skipped", "There were no go.mod or yarn.lock files to check")
+			}
+		}
+	}
+	path, err := exec.LookPath("osv-scanner")
```
academo, line 56: nit: rename to `ovsBinaryPath`

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+		pass.ReportResult(
+			pass.AnalyzerName,
+			scanningFailure,
+			"osv-scanner output not recognized",
+			fmt.Sprintf("osv-scanner output could not be parsed: %s", err))
+		return nil, nil
+	}
+
+	// no results means no issues currently reported
+	if len(objmap.Results) == 0 {
```
academo, line 107: add a report if reportall is enabled to notify this.

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+		if len(string(err.Error())) == 0 {
+			pass.ReportResult(
+				pass.AnalyzerName,
+				scanningFailure,
+				"error running osv-scanner",
+				fmt.Sprintf("osv-scanner found, but failed to run: %s", err))
+			return nil, nil
+		}
+	} else {
+		scanningFailure.Severity = analysis.OK
```
academo, line 86: only report this if `reportAll`

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,193 @@
...
+					highSeverityCount++
+				case SeverityModerate:
+					pass.ReportResult(
+						pass.AnalyzerName,
+						osvScannerModerateSeverityDetected,
+						"osv-scanner detected a moderate severity issue",
+						message)
+					moderateSeverityCount++
+				case SeverityLow:
+					pass.ReportResult(
```
academo, line 145: based on my own experience with security scanners we'll get a massive spam of low and mid security reports that will burry everything else in them.

I'd suggest we create a new config option where we can enable these reports in particular instead of reporting them by default.

pkg/analysis/passes/osvscanner/osvscanner.go
briangann: ah - i missed that - some will have both too.  this logic checks for go.mod, and if that doesn't exist, checks for yarn.lock, then proceeds with scanning. it should do both

pkg/analysis/passes/osvscanner/osvscanner.go
briangann: that's what it is doing for the reportall - if that isn't enabled yes it should be returning here

pkg/analysis/passes/osvscanner/osvscanner.go
briangann: right - after seeing the output from a variety of test files, the amount of sub-critical items can be long (and not something important enough to mitigate)

pkg/analysis/passes/osvscanner/osvscanner.go
briangann: now handles case where both exist (and one or the other)

briangann:
@academo this is working now, and also made some minor fixes to get the codebase up to current standards, now builds with mage in the workflow (and calls the linter)

pkg/analysis/passes/osvscanner/osvscanner.go
```diff
@@ -0,0 +1,222 @@
...
+		lockFile := filepath.Join(sourceCodeDir, scannerType)
+		if _, err := os.Stat(lockFile); err == nil {
+			// perform scan
+			scanningPerformed = true
+			cmdArgs := []string{"--json", "--lockfile", lockFile}
+			data, err := exec.Command(ovsBinaryPath, cmdArgs...).Output()
+			// error output is expected from osv-scanner, but if the length is zero there was a problem
+			// running the command
+			if err != nil && len(string(err.Error())) == 0 {
+				pass.ReportResult(
```
academo, line 88: I prefer with return this error instead of reporting it on the validator results. That way we literally break the validator because of something that made the validator run.

pkg/analysis/passes/osvscanner/osvscanner_test.go
```diff
@@ -0,0 +1,254 @@
...
+		ResultOf: map[*analysis.Analyzer]interface{}{
+			archive.Analyzer:    filepath.Join("testdata", "node", "critical"),
+			sourcecode.Analyzer: filepath.Join("testdata", "node", "critical"),
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+	_, err := Analyzer.Run(pass)
+	require.NoError(t, err)
+	require.Len(t, interceptor.Diagnostics, 2)
+	require.Equal(t, "osv-scanner detected a critical severity issue", interceptor.Diagnostics[0].Title)
```
academo, line 158: I had troubles with this approach (using diagnostics's indexes directly) in the past because these scanners results are not always in the same order.

Check this https://github.com/grafana/plugin-validator/blob/main/pkg/analysis/passes/jssourcemap/jssourcemap_test.go#L47

academo (approved):
thanks for all the little fixes in other files!

the osv code looks good. just a small comment in the test files to improve their robustness.

pkg/analysis/passes/osvscanner/osvscanner_test.go
briangann: very nice, i'll update the test cases to use that method. might need to have the len() to have "Greater than or Equal to" since the vulnerabilities can increase over time.

briangann: ended up using the same type of check, but with subset instead, or i'd have to enumerate all of the output.

pkg/analysis/passes/osvscanner/osvscanner.go
briangann: this now returns the error

Magefile.go
```diff
@@ -183,7 +180,7 @@ func (Build) CI(ctx context.Context) {
 func (Build) Lint() error {
 	os.Setenv("GO111MODULE", "on")
 	log.Printf("Linting...")
-	return sh.RunV("golangci-lint", "run", "./pkg/...")
+	return sh.RunV("golangci-lint", "--timeout", "2m30s", "run", "./pkg/...")
```
academo, line 183: suspiciously specific :D

briangann: CI can have issues, this made it wait long enough to finish

PR #58, opened 2023-01-23
Add basic source code validator if source code is present

academo:
Adds initial basic source code validation.

This validator will be later expanded to include checks of source code integrity with the archive file and other possible future checks.

Currently the source code validator only supports github, gitlab and bitbucket urls as well as local directories and zip files. More git services urls can later be added


pkg/analysis/analysis.go
```diff
@@ -9,10 +9,11 @@ var (
 )
 
 type Pass struct {
-	AnalyzerName string
-	RootDir      string
-	ResultOf     map[*Analyzer]interface{}
-	Report       func(string, Diagnostic)
+	AnalyzerName  string
+	RootDir       string
+	SourceCodeDir string
```
academo, line 14: Adding SourceCodeDir

pkg/repotool/repotool.go
```diff
@@ -1,31 +1,138 @@
...
 		return "", nil, err
 	}
-	_, err = git.PlainClone(tmpDir, false, &git.CloneOptions{
-		URL:      uri,
-		Progress: os.Stdout,
-		Depth:    1, // only clone the latest commit
-	})
+
+	// construct command to clone
+	cmd := []string{"git", "clone", "--depth", "1", uri, tmpDir}
```
academo, line 36: Moved from go-git to calling system command since go-git doesn't support an easy clone of a ref without specifying if it is a commit, a tag or a branch.

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/tailscale/hujson"
+
+	"github.com/grafana/plugin-validator/pkg/analysis"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/metadata"
+)
+
+var (
+	sourceCodeNotProvided      = &analysis.Rule{Name: "source-code-not-provided", Severity: analysis.Warning}
+	sourceCodeNotFound         = &analysis.Rule{Name: "source-code-not-found", Severity: analysis.Error}
+	sourceCodeVersionMissMatch = &analysis.Rule{Name: "source-code-version-missmatch", Severity: analysis.Error}
```
xnyo, line 18: Typo: this should be spelled as "mismatch", with one "s" (both in the variable name and in the `Name` string)

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -0,0 +1,87 @@
...
+	Run:      run,
+	Rules:    []*analysis.Rule{sourceCodeNotFound, sourceCodeVersionMissMatch, sourceCodeNotProvided},
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	metadataBody := pass.ResultOf[metadata.Analyzer].([]byte)
+
+	var metadata metadata.Metadata
+	if err := json.Unmarshal(metadataBody, &metadata); err != nil {
+		fmt.Println("error unmarshalling metadata")
```
xnyo, line 33: Nit: Was this left here by mistake?

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, nil
+	}
+
+	return sourceCodeDir, nil
+}
+
+func parsePackageJson(packageJsonPath string) (*PackageJson, error) {
+	rawPackageJson, err := os.ReadFile(packageJsonPath)
+	if err != nil {
+		return &PackageJson{}, err
```
xnyo, line 71: Nit: In Go, those cases are usually replaced with `return nil, err` (if the function is returning a pointer). That's shorter to write and (usually) when an error is returned the other returned value should be ignored by the caller, so `nil` is safe to return

pkg/repotool/repotool.go
```diff
@@ -1,31 +1,138 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("couldn't clone repo: %w", err)
+	}
 
 	cleanup := func() {
 		os.RemoveAll(tmpDir)
 	}
 
 	if err != nil {
+		fmt.Println("error: ", err)
```
xnyo, line 56: Nit: left here by accident as well?

xnyo (commented):
Amazing work! :clap: :clap: :clap: :clap: 
Only a couple of optional suggestions:

xnyo (approved):

pkg/analysis/passes/sourcecode/sourcecode.go
academo: Thanks! I got confused with non pointers objects which you can't return nil.

PR #60, opened 2023-01-26
Improve error messages for license checks

academo:
Improves the granularity of errors for license files checks

### DIFF
--- a/pkg/analysis/passes/license/license.go
+++ b/pkg/analysis/passes/license/license.go
@@ -1,6 +1,8 @@
 package license
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/go-enry/go-license-detector/v4/licensedb"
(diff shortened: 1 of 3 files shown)


pkg/analysis/passes/license/license.go
```diff
@@ -27,15 +29,25 @@ const minRequiredConfidenceLevel float32 = 0.9
 func run(pass *analysis.Pass) (interface{}, error) {
 	archiveDir := pass.ResultOf[archive.Analyzer].(string)
 
+	// validate that a LICENSE file is provided (go standard lib method)
+	licenseFilePath := filepath.Join(archiveDir, "LICENSE")
+	licenseFile, err := os.Stat(licenseFilePath)
+	if err != nil || licenseFile.IsDir() {
```
tolzhabayev, line 35: what is the difference between this check and the one below?

academo: They might be identical but one is done with go standard lib and the other one is done by the license library which does more checks than what I am checking here. For example the license library allows for LICENSE.md files and LICENSE sections in the readme.md file while we want an specific LICENSE file in the archive.

tolzhabayev (approved):

PR #61, opened 2023-01-30
sourcecode: Ignore checks if sourceCodeUri is not provided

xnyo:
This PR ensures the validator does not print warnings if the `sourceCodeUri` CLI argument is not provided.

This would be ideal because the workflow in our template does not have `-sourceCodeUri`:

https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/github/ci/.github/workflows/release.yml#L108

The current output is:
```
couldn't get source code: couldn't extract source code archive: . open : no such file or directory
warning: Sourcecode not provided or the provided URL does not point to a valid source code repository
detail: If you are passing a Git ref or sub-directory in the URL make sure they are correct. 
```

plugincheck2 exits with code 0, so the check does not fail, but it is confusing for the user.

Another solution would be adding `-sourceCodeUri file://.` to the plugincheck2 command inside the GitHub workflow, however that won't affect existing plugins.

Alternative solutions are welcome!

### DIFF
--- a/pkg/analysis/passes/sourcecode/sourcecode.go
+++ b/pkg/analysis/passes/sourcecode/sourcecode.go
@@ -35,7 +35,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	sourceCodeDir := pass.SourceCodeDir
 	if sourceCodeDir == "" {
-		pass.ReportResult(pass.AnalyzerName, sourceCodeNotProvided, fmt.Sprintf("Sourcecode not provided or the provided URL %s does not point to a valid source code repository", pass.SourceCodeDir), "If you are passing a Git ref or sub-directory in the URL make sure they are correct.")
+		// Silently skip check if no source code is provided, for backwards compatibility
 		return nil, nil
 	}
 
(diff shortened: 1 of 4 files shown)


marefr (approved):
Make sense to me :+1:

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -35,7 +35,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	sourceCodeDir := pass.SourceCodeDir
 	if sourceCodeDir == "" {
-		pass.ReportResult(pass.AnalyzerName, sourceCodeNotProvided, fmt.Sprintf("Sourcecode not provided or the provided URL %s does not point to a valid source code repository", pass.SourceCodeDir), "If you are passing a Git ref or sub-directory in the URL make sure they are correct.")
```
academo, line 38: Report this if reportAll is enabled.

academo (approved):

academo (approved):

PR #67, opened 2023-02-08
Fix gosec failing when no errors are returned

academo:
- Fix gosec failing with no errors
- Add debug message for source code


xnyo (approved):
:rocket:

PR #73, opened 2023-02-21
Add Published plugin validator for later validation checks

academo:
- Add published version validator and replace its usage in version validator
- This validator contains the logic to see if a plugin is published and which version
- Skip human friendly names checks if the plugin is already published

Closes https://github.com/grafana/plugin-validator/issues/72

### DIFF
--- a/pkg/analysis/passes/version/version_test.go
+++ b/pkg/analysis/passes/version/version_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/grafana/plugin-validator/pkg/analysis"
 	"github.com/grafana/plugin-validator/pkg/analysis/passes/metadata"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/published"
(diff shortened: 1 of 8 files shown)


pkg/analysis/passes/version/version_test.go
```diff
@@ -8,51 +8,17 @@ import (
...
 	"github.com/grafana/plugin-validator/pkg/analysis/passes/metadata"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/published"
 	"github.com/grafana/plugin-validator/pkg/testpassinterceptor"
 	"github.com/jarcoal/httpmock"
 	"github.com/stretchr/testify/require"
 )
 
 const testPluginId = "test-plugin-panel"
 
-func getMockVersionResponse(id string, version string) string {
```
academo, line 18: I removed all the http mocking from this test and moved to the published one. now we mock the published analyzer response directly

xnyo (approved):
LGTM!

Issue #69, opened 2023-02-10
Extend tracking url checks to use easy list tracking regex


academo:
List of domains https://blocklistproject.github.io/Lists/alt-version/tracking-nl.txt

license: `The Unlicense` at the moment of listing

PR #83, opened 2023-02-28
Use an expanded list of domains for blocked domains

academo:
- Use the tracking list for trakcing script domains

Closes https://github.com/grafana/plugin-validator/issues/69


briangann (approved):
lgtm!

PR #85, opened 2023-02-28
osv-scanner is not filtering correctly

briangann:
Plugin submissions were incorrectly being flagged as have critical vulnerabilities.

- the packages were not being filtered correctly resulting in packages like `simple-git` being included in output when it should have been excluded
- added a "common" type for noisy packages, like `underscore`
- updated critical node test yarn.lock to use an old package with a critical vulnerability that can reliably be run in the test


academo (approved):
Tested locally and it is working

PR #89, opened 2023-03-03
Add analyzer to detect standalone backend debug files

xnyo:
Backend plugins support debug mode, which is implemented by placing a `standalone.txt` file in the dist folder, which will tell Grafana to connect to a GRPC server running on the address specified in that file.

Having this file in a production plugin makes it effectively unusable.

This analyzer also checks for `pid.txt`, which will be added by https://github.com/grafana/grafana-plugin-sdk-go/pull/621 and will be used for backend debug mode (for plugins that use a new version of the SDK).


pkg/analysis/passes/backenddebug/backenddebug.go
```diff
@@ -0,0 +1,56 @@
...
+// If so, it reports an error, as the plugin can't be used properly in non-debug mode.
+var Analyzer = &analysis.Analyzer{
+	Name:     "backenddebug",
+	Requires: []*analysis.Analyzer{archive.Analyzer, metadata.Analyzer},
+	Run:      run,
+	Rules:    []*analysis.Rule{backendDebugFilePresent},
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	archiveDir := pass.ResultOf[archive.Analyzer].(string)
```
academo, line 27: `archiveDir, ok` := pass.Res and check for `ok` or `return nil,nil`. Same with the line below

pkg/analysis/passes/backenddebug/backenddebug_test.go
```diff
@@ -0,0 +1,96 @@
...
+			failureFileName: []string{},
+		},
+		{
+			name:            "standalone-txt with no executable",
+			folder:          "standalone-txt",
+			pluginJSON:      pluginJSONEmpty,
+			failureFileName: []string{},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
```
academo, line 57: I am not a fan of this kind of tests that test all conditions in a single loop. I rather have a loop for the possible failures and standalone tests for the correct case,

sympatheticmoose:
Could you add a row in https://github.com/grafana/plugin-validator#analyzers to document the analyzer? Apologies for adding new requirements mid PR!

sympatheticmoose (approved):
Thanks for adding the readme update 🥇

academo (approved):

PR #90, opened 2023-03-03
update readme

sympatheticmoose:
This PR makes a number of changes to the readme:
- General rewordings and expanding on details
- Adds CI and Go Report Shields
- Removes _arguably_ unnecessary content in the npx/local examples and strips some content from example output
- Replaces the Analyzers section with a new table format to save space and specifies the function names for each analyzer
- Removes reference for feature requests on additional git providers
- Updates npx command to add `@latest` to ensure users always run latest version and dont hit potential issues due to cached versions
- Adds new getting help section which has a generic "raise feature request" call to action
- Adds new Analyzers to the table and attempts to explain them. Some (listed below) _seemingly_ don't appear even in tool output even when `reportAll: true` is configured, so _may_ need removing? Perhaps they can't be configured individually through a config file?
  - gomanifest
  - jssourcemap
  - pluginname
  - logos
  - sourcecode
  - published (guessing just used in version?)
  - unsafesvg
- It advises that private repositories are not supported (assumption)

Potential follow ups and questions **not** addressed through this PR:
- Are local git repositories (not just zips) supported for `-sourceCodeUri`?
- Are non-git locations supported for remote zip files i.e.  FTP servers?
- Can we share or reference the Grafana Labs configuration required for publishing to the catalog (is this the default behaviour)?
- Can the naming convention be shared?
- Add (at least) code of conduct and contributing guidelines 
- The local steps will only work if the user has set up their $PATH appropriately, maybe worth advising, might be too much detail?

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ The tool runs a series of analyzers to ensure submitted plugins are following be
 | Broken Links / `brokenlinks` | Detects if any url does not resolve to a valid location. | None |
 | Code Rules / `coderules` | Checks for forbidden access to environment variables, file system or use of `syscall` module. | [semgrep](https://github.com/returntocorp/semgrep), `sourceCodeUri` |
 | Go Manifest / `gomanifest` | Validates the build manifest. | `sourceCodeUri` |
-| Go Security Checker / `gosec` | Inspects source code for security problems by scanning the Go AST. | [[gosec](https://github.com/securego/gosec)], `sourceCodeUri` |
+| Go Security Checker / `gosec` | Inspects source code for security problems by scanning the Go AST. | [gosec](https://github.com/securego/gosec), `sourceCodeUri` |
 | JS Source Map / `jssourcemap` | Checks for required module.js.map file(s) in archive. | `sourceCodeUri` |
 | Logos / `logos` | Detects whether the plugin includes small and large logos to display in the marketplace. | none |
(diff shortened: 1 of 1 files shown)


tolzhabayev (approved):
🚀

README.md
```diff
@@ -1,76 +1,84 @@
...
+Additional options can be passed to the tool:
+
 ```
 ❯ plugincheck2 -help
 Usage plugincheck2:
   -config string (optional)
         Path to configuration file
   -sourceCodeUri string (optional)
-        URI to the source code of the plugin. If set, the source code will be downloaded and analyzed. This can be a ZIP file URL, an URL to git repository or a local file (starting with `file://`)
+        URI to the source code of the plugin. If set, the source code will be downloaded and analyzed. This can be a ZIP file URL, a URL to git repository or a local file (starting with `file://`)
```
briangann, line 70: minor edit (an is correct before URL)

an URL to a git repository, or a local file (starting with `file://`)

briangann (approved):
excellent update!

README.md
sympatheticmoose: I always debate this 😄 Google suggests _a_ URL (which does read more naturally imo) - [source1]( https://www.techtarget.com/whatis/feature/Which-is-correct-a-URL-or-an-URL#:~:text=Although%20the%20U%20at%20the,use%20a%20rather%20than%20an.) [source2](https://www.quora.com/Which-is-correct-A-URL-or-An-URL)

PR #91, opened 2023-03-03
minor spelling corrections

sympatheticmoose:
Went through the various analyzers to check for misspelt words within output messages/logs.

Didn't alter id -> ID given lowercase is used in the plugin.json.


academo (approved):
Thanks!

Issue #71, opened 2023-02-15
Create an `npx` command to validate plugins


6nv:
Similar to `npx @grafana/create-plugin` or `npx @grafana/sign-plugin`, it would be very useful to have a command to validate a plugin. Then we could add it to `package.json` and run `yarn validate` (for example) to test our plugins locally before pushing to a repository.

academo:
Available here https://www.npmjs.com/package/@grafana/plugin-validator

Documentation on usage pending here https://github.com/grafana/plugin-validator/pull/88

6nv:
Great, thank you! I'm trying it out now, I think I just figured it out. These are the commands I used.

```sh
cp -r dist ventura-psychrometric-panel
zip -r ventura-psychrometric-panel.zip ventura-psychrometric-panel
npx @grafana/plugin-validator@latest -sourceCodeUri file://. ventura-psychrometric-panel.zip
```

And this is the output following:

```yml
warning: Binary for osv-scanner not found in PATH
detail: osv-scanner executable must be in your shell PATH.
error: Plugin version 1.0.0 is invalid.
detail: The submitted plugin version 1.0.0 is not greater than the latest published version 2.0.1 on grafana.com.
error: Valid license not found
detail: The provided license is not compatible with Grafana plugins. Please refer to https://grafana.com/licensing/ for more information.
warning: unsigned plugin
detail: MANIFEST.txt file not found. Please refer to the documentation for how to sign a plugin. https://grafana.com/docs/grafana/latest/developers/plugins/sign-a-plugin/
```

I'm not sure what `osv-scanner` is. Am I missing a dependency? I am on MacOSX 13.2.1 by the way.

You might want to add some steps in the README on how to run this locally. I might still not have it completely correct, so feel free to correct me.

```sh
cp -r dist ${PLUGIN_ID}
zip -r ${PLUGIN_ID}.zip ${PLUGIN_ID}
npx @grafana/plugin-validator@latest -sourceCodeUri file://. ${PLUGIN_ID}.zip
```

academo:
I updated the readme to have information about how to run the command: https://github.com/grafana/plugin-validator#with-npx

`osv-scanner` is a security tool to review your dependencies. you can install it if you want but it is optional https://github.com/google/osv-scanner

if you [run the validator with docker](https://github.com/grafana/plugin-validator#with-docker-recommended) you'll have all the security tools installed already.

6nv:
[...]

PR #92, opened 2023-03-03
Provide instruction on running the plugin validator using NPX with local files.

6nv:
See https://github.com/grafana/plugin-validator/issues/71 for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Then you can run the utility:
 plugincheck2 -sourceCodeUri [source_code_location/] [plugin_archive.zip]
 ```
 
-### NPX With Local Files
+### Generating Local Files For Validation
 
 You must create a `.zip` archive containing the `dist/` directory but named as your plugin ID.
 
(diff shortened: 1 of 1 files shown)


academo (approved):

README.md
```diff
@@ -57,6 +57,23 @@ Then you can run the utility:
 plugincheck2 -sourceCodeUri [source_code_location/] [plugin_archive.zip]
 ```
 
+### NPX With Local Files
```
sympatheticmoose, line 60: How about renaming the section: "Generating local files for validation"?

Other than the npx cmd itself, the steps - I believe - would be applicable for users of docker or locally built versions of the tool?

Also just to check - is there anything in the other plugin tools (create/sign) that would help generate the zip file?

cc @academo

6nv: > How about renaming the section: "Generating local files for validation"?

Done.

> Also just to check - is there anything in the other plugin tools (create/sign) that would help generate the zip file?

@academo should answer this one, but as far as I know, no. The `release.yml` workflow for the starter plugin uses the `mv` and `zip` commands to package the plugin. I would have assumed if there was another plugin tool to do this, it would have been used there.

academo: There's no single command to build and create the zip file.

sympatheticmoose (approved):
Thanks @6nv for your contribution ❤️

PR #96, opened 2023-03-08
Include diff report when typescript and javascript files missmatch

academo:
Adds the diff report to errors of missmatching typescript and javascript files


xnyo (approved):
LGTM!

PR #99, opened 2023-03-09
Add support for analyzer exceptions

academo:
Allows specific analyzers to be excepted for specific ids

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -19,6 +19,7 @@ trigger:
   event:
     include:
       - push
+      - pull_request
 type: docker
 ---
 kind: pipeline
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -3,7 +3,10 @@ name: Build and Test
 on:
   push:
     branches:
-      - "*"
+      - main
```
academo, line 6: Making this part of the PR as a previous mistaken PR removed the conditions for this to run

.drone.yml
```diff
@@ -19,7 +19,6 @@ trigger:
   event:
     include:
       - push
-      - pull_request
```
academo, line 22: we don't want to deploy images on PRs.

tolzhabayev (approved):

PR #103, opened 2023-03-14
osv-scanner used as a library with more reliable filtering

briangann:
This adds a revamped osv-scanner that calls the latest version (1.2.0) as a library instead of spawning a process.

Filtering has also been modified to auto-detect issues attributed to `@grafana/*` packages, without enumerating any of the details.  This is effectively a generic/automated allow list no matter which version of the grafana packages are used.

The old spawned method has been removed

- [x] Removes external dependency spawning osv-scanner
- [x] Update documentation
- [x] Remove osv-scanner binary from build process
- [x] Migrate all tests from previous method


pkg/analysis/passes/osvscanner-internal/osv_scanner_internal.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// check for the different files we check for and run scanner on each
+	// if none are detected report OK
+	scanningPerformed := false
+	for _, scannerType := range scannerTypes {
+		lockFile := filepath.Join(sourceCodeDir, scannerType)
+		if _, err := os.Stat(lockFile); err == nil {
+			// perform scan
+			scanningPerformed = true
+			data, _ := scanInternal(lockFile)
```
briangann, line 64: is there an error to detect here?

pkg/analysis/passes/osvscanner-internal/lockfile/search.go
briangann, line 1: rename file to yarnwhy

pkg/analysis/passes/osvscanner-internal/osv_scanner_internal.go
briangann, line 1: rename to osv-scanner-library vs internal

academo (approved):
Thanks for the work you put on this one @briangann !

We should follow up this with a `packageJson.lock` scanning and protection since there are plugins using npm or pnpm and we are about to introduce support for them here https://github.com/grafana/plugin-tools/pull/226

PR #116, opened 2023-04-17
Ignore SVG files for sourcemap diff

academo:
Ignores svg files to check content diff. SVG files are already verified for malicious code in another check and generally webpack will reprocess them into module imports that will keep reporting false negatives


xnyo (approved):
LGTM!

Issue #105, opened 2023-03-14
False positive matching of trackingscripts in compiled source code 


marefr:
Module.js compiled with code including 
```js
,po={};po.styleTagTransform=so()
```

And this is wrongly matching the `po.st` domain in https://github.com/grafana/plugin-validator/blob/6da5c1ca3bc71cb35be50264dfa8e1d3d3c2920e/pkg/analysis/passes/trackingscripts/list.txt#L13387

PR #120, opened 2023-04-24
Remove po.st from banned domain list

academo:
Fixes https://github.com/grafana/plugin-validator/issues/105

A more technical solution could be achieved but this is a rather unused tracking domain and it is much easier to remove it from the list than account for this specifi case


marefr (approved):
LGTM

PR #122, opened 2023-05-04
Update Angular detection patterns to include the detector script ones

xnyo:
Updates the patterns used in legacyplatform pass to match the ones in https://github.com/grafana/grafana-plugins-platform-detect-angular/blob/main/patterns.mjs


briangann (approved):
LGTM!

PR #123, opened 2023-05-16
Fix a problem with windows-generated go manifest files

academo:
Windows-generated go manifest files use windows `\` instead of `/`.

This adds the code to convert `\` to `/` when validating manifests


xnyo (approved):
LGTM!

PR #132, opened 2023-08-28
Fallback to use the first available plugin.json if not found in the expected location

academo:
- Fallbacks to detect the plugin id from the first plugin.json if not found in the expected location
- Correctly finds if the code provided in a zip file is inside a subdirectory.


briangann (approved):
looks great to me!

PR #139, opened 2023-10-05
Chore: Improve validation message when source map differ from source code

mckn:
Hopefully this will make it more obvious how the comparison is made. We got some questions from a plugin author that thought that those files was provided in the archive.


academo (approved):

sympatheticmoose (approved):
Nice

Issue #140, opened 2023-10-20
new plugin-tools creates source map files with plugin-id included in webpack path


briangann:
when using yarn v3/v4 (and any other package manager), the source map file ends up with paths like this:

```
webpack://briangann-gauge-panel/webpack/bootstrap
webpack://briangann-gauge-panel/./components/TickMaps/TickMapItem.tsx
```

the sourcediff code is only removing the leading `webpack://` which results in the tool reporting the source does not match

this was encountered using the standard plugin-tools environment but with yarn v3 in this repo https://github.com/briangann/grafana-gauge-panel/releases/tag/v2.0.0

briangann:
this doesn't appear to be limited to yarn v3, the current create-plugin setup is generating the source maps with the same paths as above

jackw:
As you've already discovered @briangann this isn't to do with package managers. Based on the logs I'm inclined to think it's due to the changes in the webpack config which enable [lazy loading](https://github.com/grafana/plugin-tools/pull/427) and prevent the additional [webpack runtimes from conflicting](https://github.com/grafana/plugin-tools/pull/464).

PR #141, opened 2023-10-20
Feature: Support arm64 and new webpack source map diffs

briangann:
- refactored getPluginId to be a utility function that can be re-used
- implements changes needed to support source maps generated by new plugin-tools webpack configuration
- updated command line tools as needed to use new params

Note: this is a general fix for the current webpack configuration and is not specific to Yarn V3.

Fixes: #140


.vscode/launch.json
```diff
@@ -30,6 +30,22 @@
...
+      "cwd": "${workspaceFolder}",
+      "args": [
+        "-nonInteractive",
+        "y",
+        "-archiveUri",
+        "https://github.com/briangann/grafana-gauge-panel/releases/download/v2.0.0/briangann-gauge-panel-2.0.0.zip",
+        "-sourceCodeUri",
+        "https://github.com/briangann/grafana-gauge-panel/tree/v2.0.0",
+      ]
+    },
```
xnyo, line 48: Nit: maybe we can also take the yarn v3 source maps from this plugin and add a separate test case for it in `pkg/analysis/passes/jssourcemap/testdata` and `pkg/analysis/passes/jssourcemap/jssourcemap_test.go`, so this new case runs in CI as well?

xnyo (commented):
Great work!! :smile: Leaving some comments:

pkg/sourcemap/sourcemap.go
```diff
@@ -30,7 +31,8 @@ var ignoreStartingWith = []string{
 
 var replaceRegex = regexp.MustCompile(`^webpack:\/*`)
 
-func ParseSourceMapFromPath(sourceMapPath string) (*sourceMap, error) {
+func ParseSourceMapFromPath(pluginID string, sourceMapPath string) (*sourceMap, error) {
+	replaceRegex = regexp.MustCompile(fmt.Sprintf("^webpack://%s/*|^webpack:/*", pluginID))
```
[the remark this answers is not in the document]
briangann: always great feedback! i'm submitting these changes now

.vscode/launch.json
briangann: i think that's a great idea, will implement it

briangann:
think we're good to go here, changes have been applied and tests are passing

PR #142, opened 2023-11-08
Update semgrep to latest

### LINKED ISSUE #135: Upgrade semgrep to latest version
In https://github.com/grafana/plugin-validator/pull/133/ we had to lock the semgrep version to 1.28 because the newer version (>=1.38) is not working as expected.

We should research what's not working in this new version and patch it so we can keep using the updated tool.

academo:
Fixes https://github.com/grafana/plugin-validator/issues/135

### DIFF
--- a/pkg/analysis/passes/coderules/coderules.go
+++ b/pkg/analysis/passes/coderules/coderules.go
@@ -75,8 +75,6 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	// run semgrep against the source code
 	semGrepArgs := []string{
 		"--json",
-		"--lang",
-		"go",
 		"--quiet",
 		"--metrics",
 		"off",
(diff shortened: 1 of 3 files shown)


pkg/analysis/passes/coderules/coderules.go
```diff
@@ -75,8 +75,6 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	// run semgrep against the source code
 	semGrepArgs := []string{
 		"--json",
-		"--lang",
```
academo, line 78: This what the problem with the update. Since the semgrep rules file already contain which language files should handle, this parameter is redundant and it was considered a "bug" for semgrep to accept it,.

briangann (approved):
lgtm, odd that it considers that a bug

PR #144, opened 2023-11-08
Add warning if using toolkit in existing plugins and errors in new plugins

### LINKED ISSUE #125: New analyzer: detect usage of Toolkit
As part of [deprecating Toolkit](https://github.com/grafana/grafana/issues/55997) we should gradually move towards rejecting plugin submissions based on this unmaintained mechanism.

Warning should be sufficient for now to further encourage migration. In or around the Grafana 10.2 timeframe we can consider changing to an error.

> Toolkit is deprecated and will not be updated to support new releases of Grafana. Please migrate to create-plugin https://grafana.github.io/plugin-tools/docs/migrating-from-toolkit.

academo:
Closes https://github.com/grafana/plugin-validator/issues/125

- Move packagejson to its own analyzer for reusability


pkg/analysis/passes/sourcecode/sourcecode_test.go
```diff
@@ -80,137 +80,3 @@ func TestSourceCodeNotProvidedReportAll(t *testing.T) {
 	require.Equal(t, "Source code not provided or the provided URL  does not point to a valid source code repository", interceptor.Diagnostics[0].Title)
```
briangann, line 80: minor item, there is an extra space here between URL and "does"

briangann (approved):
looks great, one minor text edit suggested

xnyo (approved):
LGTM, awesome! :rocket:

pkg/analysis/passes/sourcecode/sourcecode_test.go
academo: because this is a test this empty space is supposed to be the "URL". but in this test scenario there's no url because we injected the content directly.

PR #147, opened 2023-11-10
Run validators with dependencies returning `nil`

academo:
The original runner logic will skip validators that have dependencies returning nil.

This was under the idea that validator's dependencies are always required ("hard" dependencies)

Some new validators have "Soft" dependencies which means they can still run some checks even if the dependencies are nil

Additionally the code for many validators was not safe because it was optimistic about the result of the validator's dependencies and could cause panics.

This PR also fixes the code in all the validators that were optimistic about the result of the validator's dependencies.

### DIFF
--- a/pkg/analysis/passes/gomanifest/gomanifest.go
+++ b/pkg/analysis/passes/gomanifest/gomanifest.go
@@ -35,7 +35,7 @@ var Analyzer = &analysis.Analyzer{
 func run(pass *analysis.Pass) (interface{}, error) {
 	metadataBody, ok := pass.ResultOf[metadata.Analyzer].([]byte)
 	if !ok {
-		return nil, errors.New("metadata not found")
+		return nil, nil
 	}
 	var data metadata.Metadata
 	if err := json.Unmarshal(metadataBody, &data); err != nil {
--- a/pkg/analysis/passes/metadatavalid/metadatavalid.go
+++ b/pkg/analysis/passes/metadatavalid/metadatavalid.go
@@ -15,18 +15,22 @@ import (
 )
 
 var (
-	invalidMetadata = &analysis.Rule{Name: "invalid-metadata", Severity: analysis.Warning}
+	invalidMetadata  = &analysis.Rule{Name: "invalid-metadata", Severity: analysis.Warning}
+	metadataNotFound = &analysis.Rule{Name: "metadata-not-found", Severity: analysis.Error}
 )
 
 var Analyzer = &analysis.Analyzer{
 	Name:     "metadatavalid",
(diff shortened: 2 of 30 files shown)


pkg/analysis/passes/gomanifest/gomanifest.go
```diff
@@ -35,7 +35,7 @@ var Analyzer = &analysis.Analyzer{
 func run(pass *analysis.Pass) (interface{}, error) {
 	metadataBody, ok := pass.ResultOf[metadata.Analyzer].([]byte)
 	if !ok {
-		return nil, errors.New("metadata not found")
+		return nil, nil
```
academo, line 38: This change is intended. We don't want to fail the process with this. Instead we should simply skip the validator. the `metadata ` validator is already responsible of reporting this error so this is redundant logic.

pkg/analysis/passes/metadatavalid/metadatavalid.go
```diff
@@ -54,6 +61,12 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		return nil, err
 	}
 
+	_, err = os.Stat(metadataPath)
```
academo, line 64: before this would not even execute, now it executes and report a missing plugin.json

pkg/analysis/passes/modulejs/modulejs.go
```diff
@@ -24,7 +23,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	archiveDir, ok := pass.ResultOf[archive.Analyzer].(string)
 	if !ok || archiveDir == "" {
 		// this should never happen
-		return nil, fmt.Errorf("archive dir not found")
+		return nil, nil
```
academo, line 26: Intended. The error should come from the archive validator, this is redundant logic,. we can skip this validator if there's no archive

pkg/runner/runner.go
```diff
@@ -57,37 +57,30 @@ func Check(analyzers []*analysis.Analyzer, dir string, sourceCodeDir string, cfg
...
-				}
+		// run all the dependencies of the analyzer
+		for _, dep := range currentAnalyzer.Requires {
+			// if dependency returned error. This analyzer should return error too
+			if err := runFn(dep); err != nil {
+				return fmt.Errorf("%s: %w", dep.Name, err)
 			}
 		}
 
-		// TODO: Is there a better way to skip downstream analyzers than based
```
academo, line 78: This is the key change in this PR. removing this logic that skips validators with dependencies returning nil, Now all validators are run

xnyo (commented):
Great work! Just one small comment:

xnyo (approved):
LGTM great work! :partying_face:

PR #148, opened 2023-11-10
Clarify unsigned message for new plugins

academo:
New plugins get the same unsigned message as existing plugins that might be confusing for new plugin submitters.

This PR clarifies the message for new plugins

### DIFF
--- a/pkg/runner/runner_test.go
+++ b/pkg/runner/runner_test.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/grafana/plugin-validator/pkg/analysis"
 	"github.com/grafana/plugin-validator/pkg/analysis/passes"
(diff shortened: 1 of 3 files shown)


pkg/runner/runner_test.go
```diff
@@ -20,11 +20,9 @@ var tests = []struct {
 		"missing plugin.json",
 		"missing module.js",
 		"missing README.md",
-		"unsigned plugin",
```
academo, line 23: It is odd that these tests are different because of the manifest changes but it is related to dependencies and how the runner runs them. 

This PR https://github.com/grafana/plugin-validator/pull/147 is trying to fix these inconsistencies.

xnyo (approved):
LGTM! :smiley:

PR #166, opened 2024-02-19
Add checksum validation 

academo:
Add checksum validator with cli parameter

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -18,7 +18,7 @@ var (
 type Pass struct {
 	AnalyzerName string
 	RootDir      string
-	CheckParams  *CheckParams
+	CheckParams  CheckParams
 	ResultOf     map[*Analyzer]interface{}
 	Report       func(string, Diagnostic)
 }
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -16,12 +16,19 @@ var (
 )
 
 type Pass struct {
-	AnalyzerName  string
-	RootDir       string
-	SourceCodeDir string
-	Checksum      string
-	ResultOf      map[*Analyzer]interface{}
-	Report        func(string, Diagnostic)
+	AnalyzerName string
+	RootDir      string
+	CheckParams  *CheckParams
+	ResultOf     map[*Analyzer]interface{}
+	Report       func(string, Diagnostic)
+}
+
+type CheckParams struct {
+	ArchiveDir            string
+	SourceCodeDir         string
+	Checksum              string
+	ArchiveCalculatedMD5  string
+	ArchiveCalculatedSHA1 string
 }
 
 func (p *Pass) ReportResult(analysisName string, rule *Rule, message string, detail string) {
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -19,6 +19,7 @@ type Pass struct {
 	AnalyzerName  string
 	RootDir       string
 	SourceCodeDir string
+	Checksum      string
 	ResultOf      map[*Analyzer]interface{}
 	Report        func(string, Diagnostic)
 }
--- a/pkg/analysis/passes/checksum/checksum.go
+++ b/pkg/analysis/passes/checksum/checksum.go
@@ -1,0 +1,155 @@
+package checksum
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/grafana/plugin-validator/pkg/analysis"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/readme"
+	"github.com/grafana/plugin-validator/pkg/logme"
+)
+
+var (
+	checksumInvalid     = &analysis.Rule{Name: "checksum-invalid", Severity: analysis.Error}
+	checksumFormatError = &analysis.Rule{Name: "checksum-format-error", Severity: analysis.Error}
+)
+
+var Analyzer = &analysis.Analyzer{
+	Name:     "checksum",
+	Requires: []*analysis.Analyzer{readme.Analyzer},
+	Run:      run,
+	Rules:    []*analysis.Rule{checksumInvalid},
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+
+	checksum := pass.CheckParams.Checksum
+
+	// skip if no provided
+	if checksum == "" || len(checksum) == 0 {
+		return nil, nil
+	}
+
+	checksum, err := resolveCheckSum(checksum)
+	if err != nil {
+		pass.ReportResult(
+			pass.AnalyzerName,
+			checksumFormatError,
+			fmt.Sprintf("%s: %s", err.Error(), pass.CheckParams.Checksum),
+			"Make sure you provide a valid checksum as a string or a valid URL to a checksum. Only MD5 and SHA1 checksums are supported.",
+		)
+		return nil, nil
+	}
+
+	checkSumType := ""
+
+	// determine if it is md5 or sha1
+	if len(checksum) == 32 {
+		checkSumType = "md5"
+	} else if len(checksum) == 40 {
+		checkSumType = "sha1"
+	} else {
+		pass.ReportResult(
+			pass.AnalyzerName,
+			checksumFormatError,
+			fmt.Sprintf("Invalid checksum format: %s", checksum),
+			"Make sure you provide a valid checksum as a string or a valid URL to a checksum. Only MD5 and SHA1 checksums are supported.",
+		)
+		return nil, nil
+	}
+	isError := false
+
+	switch checkSumType {
+	case "md5":
+		if checksum != pass.CheckParams.ArchiveCalculatedMD5 {
+			isError = true
+		}
+	case "sha1":
+		if checksum != pass.CheckParams.ArchiveCalculatedSHA1 {
+			isError = true
+		}
+	}
+
+	if isError {
+		pass.ReportResult(
+			pass.AnalyzerName,
+			checksumInvalid,
+			fmt.Sprintf("The provided checksum %s does not match the plugin archive", checksum),
+			"The plugin archive does not match the provided checksum in the submission form. Please double check you are providing the correct plugin archive and checksum.",
+		)
+	}
+
+	return nil, nil
+}
+
+// sums can come as:
+// 04b2f189616b65985d219ca094f3609b  arduino-cli.yaml
+// 0b41e5bffba89764a494f420347fe5110fa2886c  arduino-cli.yaml
+// 0b41e5bffba89764a494f420347fe5110fa2886c
+// 04b2f189616b65985d219ca094f3609b
+// we only care about the first part
+func sanitizeCheckSum(checksum string) string {
+	return strings.Split(checksum, " ")[0]
+}
+
+// checksum can be urls, we need to download
+func resolveCheckSum(checksum string) (string, error) {
+
+	if strings.HasPrefix(checksum, "https://") || strings.HasPrefix(checksum, "http://") {
+		finalChecksum, err := getFirstLineFromURL(checksum)
+		if err != nil {
+			return "", err
+		}
+		return sanitizeCheckSum(finalChecksum), nil
+	}
+	return sanitizeCheckSum(checksum), nil
+}
+
+const (
+	maxSize = 500 * 1024 // 500KB
+	timeout = 30 * time.Second
+)
+
+func getFirstLineFromURL(url string) (string, error) {
+	client := http.Client{
+		Timeout: timeout,
+	}
+
+	resp, err := client.Get(url)
+	if err != nil {
+		logme.DebugFln("Error reading body: %s", err.Error())
+		return "", err
+	}
+
+	if resp.StatusCode != 200 {
+		return "", errors.New("checksum file not found")
+	}
+
+	defer resp.Body.Close()
+
+	if resp.ContentLength > maxSize {
+		return "", errors.New("checksum file is too large")
+	}
+
+	lr := &io.LimitedReader{R: resp.Body, N: maxSize}
+	body, err := io.ReadAll(lr)
+	if err != nil {
+		logme.DebugFln("Error reading body: %s", err.Error())
+		return "", err
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(body)))
+	scanner.Scan() // get the first line
+	line := scanner.Text()
+
+	if err := scanner.Err(); err != nil {
+		return "", err
+	}
+
+	return line, nil
+}
--- a/pkg/analysis/passes/checksum/checksum.go
+++ b/pkg/analysis/passes/checksum/checksum.go
@@ -35,7 +35,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		return nil, nil
 	}
 
-	checksum, err := resolveCheckSum(checksum)
+	checksum, err := resolveCheckSum(strings.TrimSpace(checksum))
 	if err != nil {
 		pass.ReportResult(
 			pass.AnalyzerName,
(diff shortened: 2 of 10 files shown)


pkg/analysis/analysis.go
```diff
@@ -16,11 +16,19 @@ var (
 )
 
 type Pass struct {
-	AnalyzerName  string
-	RootDir       string
-	SourceCodeDir string
-	ResultOf      map[*Analyzer]interface{}
-	Report        func(string, Diagnostic)
+	AnalyzerName string
```
academo, line 19: Since we are passing more global data to the analyzers I decided to move these parameters to its own object instead of having this function positional parameters growing ever more.

This required some minor changes in other test files and the runner but the functionality remains untouched

pkg/analysis/passes/packagejson/packagejson_test.go
```diff
@@ -24,8 +24,10 @@ func TestVersionSourceCodeMatch(t *testing.T) {
 	}`)
 
 	pass := &analysis.Pass{
-		RootDir:       filepath.Join("./"),
-		SourceCodeDir: filepath.Join("testdata", "version-match"),
+		RootDir: filepath.Join("./"),
+		CheckParams: &analysis.CheckParams{
```
academo, line 28: Changing this since we moved to an object instead of positional parameters

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -8,25 +8,46 @@ import (
...
 
 func run(pass *analysis.Pass) (interface{}, error) {
 
-	sourceCodeDir := pass.SourceCodeDir
+	sourceCodeDir := pass.CheckParams.SourceCodeDir
 	if sourceCodeDir == "" {
 		// If no source code dir is provided, only report the result if ReportAll is set, for backwards compatibility
 		if sourceCodeNotProvided.ReportAll {
-			pass.ReportResult(pass.AnalyzerName, sourceCodeNotProvided, fmt.Sprintf("Source code not provided or the provided URL %s does not point to a valid source code repository", pass.SourceCodeDir), "If you are passing a Git ref or sub-directory in the URL make sure they are correct.")
+			pass.ReportResult(
```
academo, line 42: breaking down these long lines for easier reading. no changes in text.

pkg/cmd/plugincheck2/main.go
```diff
@@ -76,7 +104,17 @@ func main() {
 		defer sourceCodeDirCleanup()
 	}
 
-	diags, err := runner.Check(passes.Analyzers, archiveDir, sourceCodeDir, cfg)
+	diags, err := runner.Check(
+		passes.Analyzers,
+		&analysis.CheckParams{
```
academo, line 109: This is the change from positional parameters to an object with global data we can pass to the analyzers.

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// sums can come as:
+// 04b2f189616b65985d219ca094f3609b  arduino-cli.yaml
+// 0b41e5bffba89764a494f420347fe5110fa2886c  arduino-cli.yaml
+// 0b41e5bffba89764a494f420347fe5110fa2886c
+// 04b2f189616b65985d219ca094f3609b
+// we only care about the first part
+func sanitizeCheckSum(checksum string) string {
+	return strings.Split(checksum, " ")[0]
```
xnyo, line 97: Can they also be separated by other whitespaces (like tabs) or just spaces? In case there can also be other whitespace characters, you can use [strings.Fields](https://pkg.go.dev/strings#Fields) instead of strings.Split with a space

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	return sanitizeCheckSum(checksum), nil
+}
+
+const (
+	maxSize = 500 * 1024 // 500KB
+	timeout = 30 * time.Second
+)
+
+func getFirstLineFromURL(url string) (string, error) {
```
xnyo, line 118: Nit: Ideally we would pass a `context.Context` here since we are making a network request, but I noticed we aren't doing so also in other parts of the code already. I can open an issue and add `context.Context` where it makes sense.

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+	client := http.Client{
+		Timeout: timeout,
+	}
+
+	resp, err := client.Get(url)
+	if err != nil {
+		logme.DebugFln("Error reading body: %s", err.Error())
```
xnyo, line 125: Nit: this is an error making the network request not while reading the body so I would change the logged message to something more clear

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+		return "", err
+	}
+
+	if resp.StatusCode != 200 {
+		return "", errors.New("checksum file not found")
```
xnyo, line 130: Status code != 200 could also be another reason, not necessarily file not found. I would also add the original status code to the error, something like:

```suggestion
		return "", fmt.Errorf("bad status code: %d", resp.StatusCode)
```

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	defer resp.Body.Close()
```
xnyo, line 133: This should be done as soon as the `err` returned by `client.Get` is `nil`, so also in case the status code is not 200. This needs to be moved before the `if resp.StatusCode != 200` check above

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	lr := &io.LimitedReader{R: resp.Body, N: maxSize}
+	body, err := io.ReadAll(lr)
+	if err != nil {
+		logme.DebugFln("Error reading body: %s", err.Error())
+		return "", err
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(body)))
```
xnyo, line 146: `io.LimitedReader` implements `io.Reader`, this means you should be able to pass it directly to `bufio.NewScanner`, like this:

```suggestion
	lr := &io.LimitedReader{R: resp.Body, N: maxSize}
	scanner := bufio.NewScanner(lr)
```

pkg/analysis/passes/checksum/checksum.go
```diff
@@ -0,0 +1,155 @@
...
+	switch checkSumType {
+	case "md5":
+		if checksum != pass.CheckParams.ArchiveCalculatedMD5 {
+			isError = true
+		}
+	case "sha1":
+		if checksum != pass.CheckParams.ArchiveCalculatedSHA1 {
+			isError = true
+		}
+	}
```
xnyo, line 76: Nit: Maybe make the check lower case? Something like:

```suggestion
	switch checkSumType {
	case "md5":
		if strings.ToLower(checksum) != strings.ToLower(pass.CheckParams.ArchiveCalculatedMD5) {
			isError = true
		}
	case "sha1":
		if strings.ToLower(checksum) != strings.ToLower(pass.CheckParams.ArchiveCalculatedSHA1) {
			isError = true
		}
	}
```

pkg/analysis/passes/checksum/checksum_test.go
```diff
@@ -0,0 +1,277 @@
...
+			ArchiveCalculatedSHA1: "187cd80948f240957bf8399745335b89f005c5f0",
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "https://example.com/checksum",
+		httpmock.NewStringResponder(200, "f9b1c42c45cbf4953d7da5c31b3d73d9 "))
```
xnyo, line 156: Nit: Lint :smiley: 

```suggestion
	httpmock.RegisterResponder("GET", "https://example.com/checksum",
	httpmock.NewStringResponder(200, "f9b1c42c45cbf4953d7da5c31b3d73d9 "))
```

(we should probably add a `golangci-lint` step to the CI pipeline for PRs in this repo, I can open an issue for that)

pkg/analysis/passes/checksum/checksum_test.go
```diff
@@ -0,0 +1,277 @@
...
+			ArchiveCalculatedSHA1: "187cd80948f240957bf8399745335b89f005c5f0",
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "https://example.com/checksum",
+		httpmock.NewStringResponder(200, "f9b1c42c45cbf4953d7da5c31b3d73d9 archive-file-name.zip"))
```
xnyo, line 181:
```suggestion
	httpmock.RegisterResponder("GET", "https://example.com/checksum",
	httpmock.NewStringResponder(200, "f9b1c42c45cbf4953d7da5c31b3d73d9 archive-file-name.zip"))
```

pkg/analysis/passes/checksum/checksum_test.go
```diff
@@ -0,0 +1,277 @@
...
+			ArchiveCalculatedSHA1: "187cd80948f240957bf8399745335b89f005c5f0",
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "https://example.com/checksum",
+		httpmock.NewStringResponder(200, "x9b1c42c45cbf4953d7da5c31b3d73d3"))
```
xnyo, line 206:
```suggestion
	httpmock.RegisterResponder("GET", "https://example.com/checksum",
	httpmock.NewStringResponder(200, "x9b1c42c45cbf4953d7da5c31b3d73d3"))
```

pkg/analysis/passes/checksum/checksum_test.go
```diff
@@ -0,0 +1,277 @@
...
+			ArchiveCalculatedSHA1: "187cd80948f240957bf8399745335b89f005c5f0",
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "https://example.com/checksum",
+		httpmock.NewStringResponder(404, ""))
```
xnyo, line 236:
```suggestion
	httpmock.RegisterResponder("GET", "https://example.com/checksum",
	httpmock.NewStringResponder(404, ""))
```

pkg/analysis/passes/checksum/checksum_test.go
```diff
@@ -0,0 +1,277 @@
...
+			ArchiveCalculatedSHA1: "187cd80948f240957bf8399745335b89f005c5f0",
+		},
+		Report: interceptor.ReportInterceptor(),
+	}
+
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	httpmock.RegisterResponder("GET", "https://example.com/checksum",
+		httpmock.NewStringResponder(200, "{\"test\": \"test\"}"))
```
xnyo, line 266:
```suggestion
	httpmock.RegisterResponder("GET", "https://example.com/checksum",
	httpmock.NewStringResponder(200, "{\"test\": \"test\"}"))
```

pkg/analysis/analysis.go
```diff
@@ -16,11 +16,19 @@ var (
...
+	RootDir      string
+	CheckParams  *CheckParams
```
xnyo, line 21: Can/does it make sense for `CheckParams` to be `nil`? I don't think it would make sense, in that case I would change this from a `*CheckParams` (pointer) to a `CheckParams` (value). This way it cannot be nil and it cannot panic if trying to access its attributes

pkg/analysis/passes/sourcecode/sourcecode.go
```diff
@@ -8,25 +8,46 @@ import (
...
+		sourceCodeNotFound,
+		sourceCodeVersionMisMatch,
+		sourceCodeNotProvided,
+	},
 }
 
 func run(pass *analysis.Pass) (interface{}, error) {
 
-	sourceCodeDir := pass.SourceCodeDir
+	sourceCodeDir := pass.CheckParams.SourceCodeDir
```
xnyo, line 38: `pass.CheckParams` could potentially be `nil`. I would probably change `pass.CheckParams` to be a value rather than a pointer

pkg/cmd/plugincheck2/main.go
```diff
@@ -28,17 +30,32 @@ type FormattedOutput struct {
...
-		sourceCodeUri = flag.String("sourceCodeUri", "", "URL to the source code of the plugin. If set, the source code will be downloaded and analyzed. This can be a ZIP file or an URL to git repository")
+		sourceCodeUri = flag.String(
+			"sourceCodeUri",
+			"",
+			"URL to the source code of the plugin. If set, the source code will be downloaded and analyzed. This can be a ZIP file or an URL to git repository",
+		)
+		checksum = flag.String(
+			"checksum",
+			"",
+			"checksum of the plugin archive. MD5 or SHA1. A string with the hash or an url to a file with the hash",
```
xnyo, line 47: Nit: I would make this a bit more clear:

```suggestion
			"checksum of the plugin archive. MD5, SHA1 or a string with the hash or an url to a file with the hash",
```

xnyo (commented):
Great work! :clap: :clap: 

I have some comments, some of those like missing `context.Context` are general issues with the code of this repository, but I wanted to point them out anyways.

Apart from those, great work!! :tada:

pkg/analysis/passes/checksum/checksum.go
academo: As far as I know it can be only whitespaces but I like the idea of using something that can handle them all.

pkg/analysis/passes/checksum/checksum.go
academo: @xnyo it is indeed. but this is an error message that the user will read (in the submission process) not to log somewhere for us so I decided to make it somewhat more user-readable and if it comes to it we can debug the issue further if it ever happens.

pkg/analysis/passes/checksum/checksum.go
academo: great catch!

pkg/analysis/passes/checksum/checksum.go
xnyo: Okay, sounds good then!

pkg/analysis/passes/checksum/checksum_test.go
academo: this is not an indentation error but intended, this is a function parameter split in two lines so the indentation exists to show that they are part of the same line. (this is auto-formatted by https://github.com/segmentio/golines)

pkg/analysis/passes/checksum/checksum_test.go
academo: Likewise. this is not an indentation error but autoformatted by https://github.com/segmentio/golines

pkg/analysis/passes/checksum/checksum_test.go
xnyo: My bad :facepalm:

pkg/analysis/passes/checksum/checksum_test.go
academo: This is part of https://github.com/segmentio/golines to make long lines more readable. All the code is already well formatted with gofmt too

pkg/analysis/passes/sourcecode/sourcecode.go
academo: I changed it to a value!

xnyo (approved):
LGTM! Awesome :tada:

PR #169, opened 2024-03-01
Fix checksum handling of wrong urls

academo:
Fix an error when the url returns empty content. This can happen if the provided url is to a download service or a page to see the checksum instead of a raw checksum


briangann (approved):
lgtm

xnyo (approved):
LGTM! :tada:

PR #172, opened 2024-03-15
Angular detection: Fetch patterns from GCOM

### LINKED ISSUE #137: Angular detection: Fetch patterns from GCOM
In the `legacyplatform` inspector (Angular), it would be good to fetch the patterns from GCOM rather than relying on hardcoded ones:

GCOM API: https://grafana.com/api/plugins/angular_patterns

https://github.com/grafana/plugin-validator/blob/0b03aac7dea53a1a0bab4d737aac7bbfb621ff32/pkg/analysis/passes/legacyplatform/legacyplatform.go#L49-L62

Grafana core already has support for it:

https://github.com/grafana/grafana/blob/e9a12598db238b3923ca30a490d4abcaf0b00b65/pkg/services/pluginsintegration/angulardetectorsprovider/dynamic.go#L110

This would save us from having to open yet another PR when we have to chang the Angular detection patterns

oshirohugo:
## What is this PR about
Fetch patterns from angular detection from GCOM and add simple tests

## Note for reviewers
Not all patterns that were hardcoded are returned by gcom, therefore I only replaced the ones that are returned.
Also the `QueryCtrl` detection is now a little bit different: it's based on a regex and will detect: `"QueryCtrl"` and `'QueryCtrl'`. Before, the hardcoded version was only looking if the `QueryCtrl` was contained.

Closes https://github.com/grafana/plugin-validator/issues/137

### DIFF
--- a/pkg/analysis/passes/legacyplatform/legacyplatform.go
+++ b/pkg/analysis/passes/legacyplatform/legacyplatform.go
@@ -2,6 +2,9 @@ package legacyplatform
 
 import (
 	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
 	"regexp"
 
 	"github.com/grafana/plugin-validator/pkg/analysis"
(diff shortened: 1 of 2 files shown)


pkg/analysis/passes/legacyplatform/legacyplatform.go
```diff
@@ -47,19 +50,52 @@ func (d *regexDetector) Detect(moduleJs []byte) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var patterns []gcomPattern
+
+	err = json.Unmarshal(body, &patterns)
+	if err != nil {
+		return nil, err
+	}
```
xnyo, line 83: Nit:
```suggestion
	var patterns []gcomPattern
	if err := json.NewDecoder(resp.Body).Decode(patterns); err != nil {
		return nil, err
	}
```

pkg/analysis/passes/legacyplatform/legacyplatform.go
```diff
@@ -47,19 +50,52 @@ func (d *regexDetector) Detect(moduleJs []byte) bool {
...
-	&containsBytesDetector{pattern: []byte("app/plugins/sdk")},
-	&containsBytesDetector{pattern: []byte("angular.isNumber(")},
-	&containsBytesDetector{pattern: []byte("editor.html")},
-	&containsBytesDetector{pattern: []byte("ctrl.annotation")},
 	&containsBytesDetector{pattern: []byte("System.register(")},
 
-	// &regexDetector{regex: regexp.MustCompile(`['"](app/core/.*?)|(app/plugins/.*?)['"]`)},
+	&regexDetector{regex: regexp.MustCompile(`['"](app/core/.*?)|(app/plugins/.*?)['"]`)},
 	&regexDetector{regex: regexp.MustCompile(`['"](app/core/utils/promiseToDigest)|(app/plugins/.*?)|(app/core/core_module)['"]`)},
 	&regexDetector{regex: regexp.MustCompile(`from\s+['"]grafana\/app\/`)},
```
xnyo, line 57: I *think* we should remove this `legacyDetectors` altogether and use only GCOM as the only source of truth for angular detections.

In case there are patterns here that are not in GCOM API, then we have removed them from GCOM because they were redundant.

xnyo (commented):
Awesome work!! 👏👏👏 Thanks for taking care of this!

Just a couple of small comments:

xnyo (approved):
LGTM, awesome work! :clap: :clap: :clap:

PR #179, opened 2024-04-02
Fix flaky tests when legacyplatform analyzer is invoked

xnyo:
Fixes flaky tests involving the `legacyplatform` analyzer (angular) when `https://grafana.com/api/plugins/angular_patterns` is called too quickly:

```
--- FAIL: TestRunner (0.39s)
    --- FAIL: TestRunner/AllFilesPresentButEmpty (0.23s)
        runner_test.go:72: legacyplatform: json: cannot unmarshal object into Go value of type []legacyplatform.gcomPattern
FAIL
FAIL    github.com/grafana/plugin-validator/pkg/runner  0.488s

```

This is happening because calling the API too fast will trigger the rate limit in GCOM:
```
{
  "code": "TooManyRequests",
  "message": "Rate limit exceeded.",
  "requestId": "d81fb3f1-ed30-4e66-8da6-4dc7fb7a2d38"
}
```

With this PR, we cache the result and call the GCOM API only once

### DIFF
--- a/pkg/analysis/passes/legacyplatform/legacyplatform.go
+++ b/pkg/analysis/passes/legacyplatform/legacyplatform.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"regexp"
+	"sync"
 
 	"github.com/grafana/plugin-validator/pkg/analysis"
 	"github.com/grafana/plugin-validator/pkg/analysis/passes/modulejs"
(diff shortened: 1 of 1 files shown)


andresmgot (commented):
Do we need to expire the cache?

xnyo:
> Do we need to expire the cache?

No, those hardly ever change and it this only lives in memory anyways, so re-running the plugin-validator will expire the cache.

We only need it so we don't hit GCOM API too quickly and get back a 429, mostly when running tests

andresmgot (approved):
LGTM

PR #186, opened 2024-04-25
Validate all files have at least read permissions

### LINKED ISSUE #185: Verify file permissions in zip files
Even though we do our best to try to set the file permissions correctly when we generate them - it can happen that sometimes those file permissions get modified by users.

Validator should have a new check to ensure that the permissions are set properly also for other files than executables https://github.com/grafana/plugin-validator/blob/18ec4dc7c75b222e3641de979936159a1cc86b82/pkg/analysis/passes/binarypermissions/binarypermissions.go#L4

### Check that the following permissions are set:
- At least 640? on all files in the zip file for signature verification

Example of where it failed recently was plugin manifest file  called `go_plugin_build_manifest` which is usually generated with `755`
https://github.com/grafana/grafana-plugin-sdk-go/blob/3bbfa89bed9c0882140eee5004e2657cf9ae2fce/build/common.go#L205

But for some reason ended up being `600` in the following plugin https://grafana.com/api/plugins/oci-metrics-datasource/versions/5.2.0/download

This led the `verfiyHash` function to fail and mark plugin as invalid signature / modified signature: https://github.com/grafana/grafana/blob/0220fd588a2ef308ede052da79d5daa91eb33015/pkg/plugins/manager/signature/manifest.go#L264 on Grafana Cloud. We did not catch this locally as we are running Grafana probably with different set of privileges and did not have this issue.

academo:
- Add validation to check that files in the manifest have at least read permissions. Note that there is a separate validation step to check binary permissions
- Formatted the files with gofmt, goimports and golines for more readability

Closes https://github.com/grafana/plugin-validator/issues/185


pkg/analysis/passes/manifest/manifest.go
```diff
@@ -9,13 +9,13 @@ import (
 	"path/filepath"
```
academo, line 9: note: format applied.

pkg/analysis/passes/manifest/manifest.go
```diff
@@ -104,19 +136,52 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+			pass.ReportResult(
+				pass.AnalyzerName,
+				invalidShaSum,
+				"invalid file checksum",
+				fmt.Sprintf("checksum for file %s is invalid", relativePath),
+			)
+			return nil
+		}
+
+		mode := file.Mode()
```
academo, line 170: This is the relevant change.

pkg/analysis/passes/manifest/manifest_test.go
```diff
@@ -173,6 +182,39 @@ func TestCommunityWithRootUrls(t *testing.T) {
...
+		"MANIFEST.txt: plugin signature contains rootUrls",
+	)
+	require.Equal(
+		t,
+		interceptor.Diagnostics[0].Detail,
+		"The plugin is signed as community but contains rootUrls. Do not pass --rootUrls when signing this plugin as community type",
+	)
+}
+
+func TestWithWrongPermissions(t *testing.T) {
```
academo, line 197: This is the relevant new test.

pkg/analysis/passes/manifest/testdata/with-wrong-permissions/CHANGELOG.md
```diff
@@ -0,0 +1,62 @@
+# Change Log
```
academo, line 1: Ignore all the files from here to the end. they are fixtures for the test

marefr (approved):
LGTM

PR #192, opened 2024-05-06
New Validator: Add sdk usage validation for plugins with a backend component

academo:
Adds the code to validate if the plugin has a backend component it should use the grafana go sdk

Note: Currently not taking in consideration nested plugins. That is to come in a follow up PR


marefr (approved):
LGTM

PR #173, opened 2024-03-20
chore: update packages and images

briangann:
* Use node 20 (was 16)
* Use securego v2.19.0 (was v2.18.2)
* Use golangci-lint v1.57.1 (was 1.55.2)
* Use golang v1.22 (was 1.21)
* Updates all go modules
* Bump base image to alpine:3.19

PR #195, opened 2024-05-07
Add integration tests for the validator

academo:
- **Add integration tests**
- **run integration tests on workflow**

These current integration tests are far from a complete test case of the validator and more will need to be added.


.github/workflows/test.yml
```diff
@@ -14,24 +14,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Go
```
academo, line 17: These were added back by mistake here https://github.com/grafana/plugin-validator/pull/173 after they were removed by https://github.com/grafana/plugin-validator/pull/181

Currently these don't do anything

pkg/integrationtests/integrationtests.go
```diff
@@ -0,0 +1,233 @@
...
+	Name     string `json:"name"`
+}
+
+type JsonReport struct {
+	Id              string             `json:"id"`
+	Version         string             `json:"version"`
+	PluginValidator map[string][]Issue `json:"plugin-validator"`
+}
+
+var files = map[string]JsonReport{
```
academo, line 27: effectively the integration tests will run for each of the files defined in this map and compare that the validator outputs the same errors and warnings. It'll exit 1 if any of the keys changed.

Currently the reporting is not optimal as it is just a diff report but it will be able to detect changes on deep changes inside the validator.

Magefile.go
```diff
@@ -285,3 +298,9 @@ func (Run) SourceDiffLocal(ctx context.Context, archive string, source string) e
 	return sh.RunV(command[0], command[1:]...)
 
 }
+
+func (Run) IntegrationTests() error {
```
xnyo, line 302: I would move this target under the `Test` target, and also add a docstring so it's displayed when running `mage -l`,  as well as moving this up next to the other `(Test)` targets:

```suggestion
// Integration runs the integration tests for the plugin validator
func (Test) Integration() error {
```


This way it can be invoked via

```bash
mage test:integration
```

Which I think is more suited. WDYT?

pkg/integrationtests/integrationtests.go
xnyo, line 1: Awesome work on this! :partying_face: 

This looks really good, but I would change it so it uses the go testing framework intsead. This way it's easier to write assertions for it (you can use testify so `require` and `assert`), and it will also give you a report about tests passing or failing like any other go test. For example:

```
╰─❯ mage run:integrationTests
--- FAIL: TestIntegration (0.00s)
    --- FAIL: TestIntegration/invalid.zip (0.28s)
        main_test.go:187: Running invalid.zip
        main_test.go:195: Running command: bin/linux_amd64/plugincheck2 -config config/integration-tests.yaml pkg/cmd/plugincheck2/testdata/invalid.zip
        main_test.go:202: 
                Error Trace:    /home/giuseppe/grafana/plugin-validator/pkg/cmd/plugincheck2/main_test.go:202
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestIntegration/invalid.zip
    --- FAIL: TestIntegration/invalid2.zip (1.86s)
        main_test.go:187: Running invalid2.zip
        main_test.go:195: Running command: bin/linux_amd64/plugincheck2 -config config/integration-tests.yaml pkg/cmd/plugincheck2/testdata/invalid2.zip
        main_test.go:202: 
                Error Trace:    /home/giuseppe/grafana/plugin-validator/pkg/cmd/plugincheck2/main_test.go:202
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestIntegration/invalid2.zip
FAIL
FAIL    github.com/grafana/plugin-validator/pkg/cmd/plugincheck2        6.996s
FAIL
Error: running "go test ./pkg/cmd/plugincheck2" failed with exit code 1
```

or:

```
╭─ ~/grafana/plugin-validator on academo/integration-tests +2 !4 ?13                                                        took 10s
╰─❯ mage test:integration
=== RUN   TestIntegration
=== RUN   TestIntegration/grafana-clock-panel-2.1.5.any.zip
=== PAUSE TestIntegration/grafana-clock-panel-2.1.5.any.zip
=== RUN   TestIntegration/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
=== PAUSE TestIntegration/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
=== RUN   TestIntegration/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
=== PAUSE TestIntegration/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
=== RUN   TestIntegration/invalid.zip
=== PAUSE TestIntegration/invalid.zip
=== RUN   TestIntegration/invalid2.zip
=== PAUSE TestIntegration/invalid2.zip
=== CONT  TestIntegration/grafana-clock-panel-2.1.5.any.zip
    main_test.go:204: Running grafana-clock-panel-2.1.5.any.zip
    main_test.go:212: Running command: ../../../bin/linux_amd64/plugincheck2 -config testdata/config/integration-tests.yaml testdata/grafana-clock-panel-2.1.5.any.zip
=== CONT  TestIntegration/invalid.zip
=== CONT  TestIntegration/invalid2.zip
    main_test.go:204: Running invalid2.zip
    main_test.go:212: Running command: ../../../bin/linux_amd64/plugincheck2 -config testdata/config/integration-tests.yaml testdata/invalid2.zip
=== CONT  TestIntegration/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
    main_test.go:204: Running yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
=== CONT  TestIntegration/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
=== NAME  TestIntegration/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
    main_test.go:212: Running command: ../../../bin/linux_amd64/plugincheck2 -config testdata/config/integration-tests.yaml testdata/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip
=== NAME  TestIntegration/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
    main_test.go:204: Running alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
    main_test.go:212: Running command: ../../../bin/linux_amd64/plugincheck2 -config testdata/config/integration-tests.yaml testdata/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip
=== NAME  TestIntegration/invalid.zip
    main_test.go:204: Running invalid.zip
    main_test.go:212: Running command: ../../../bin/linux_amd64/plugincheck2 -config testdata/config/integration-tests.yaml testdata/invalid.zip
--- PASS: TestIntegration (0.00s)
    --- PASS: TestIntegration/invalid.zip (0.14s)
    --- PASS: TestIntegration/invalid2.zip (1.76s)
    --- PASS: TestIntegration/grafana-clock-panel-2.1.5.any.zip (2.85s)
    --- PASS: TestIntegration/alexanderzobnin-zabbix-app-4.4.9.linux_amd64.zip (3.18s)
    --- PASS: TestIntegration/yesoreyeram-infinity-datasource-2.6.3.linux_amd64.zip (6.77s)
PASS
ok      github.com/grafana/plugin-validator/pkg/cmd/plugincheck2        6.844s
```

To achieve this, I would do the following:

- Instead of having a `pkg/integrationtests`, I would declare the integration tests in `pkg/cmd/plugincheck2/main_test.go`. In my mind this makes sense because they test the whole program (so the `pkg/cmd/plugincheck2`).
- Change the `package` directive at the top of the file with `package main_test`
- Move the zip files from `testdata/*.zip` into `pkg/cmd/plugincheck2/testdata/*.zip`, so they are next to the `_test.go` file. I would also move the config into `pkg/cmd/plugincheck2/testdata/config/integration-tests.yaml` to keep everything in one place
- Instead of having a `func RunTests()`, declare a `func TestIntegration(t *testing.T)`. This way the function can be run via `go test` like any other go test

The majority of the file can stay the same, but I will also make the following changes:
- Use `t.Run()` to declare each test case as its own, and use `t.Parallel()` at the top of each test case so they run in parallel and run faster
- Use `t.Log` instead of `fmt.Printf`/`fmt.Println`, so the text is logged alongside the test case name
- Use `require` and `assert` instead of an if with a return

High level-overview of those changes:
```go

func TestIntegration(t *testing.T) {
	// (weird path because those now run in pkg/cmd/plugincheck2, and the binary is built in bin)
	binary := filepath.Join("..", "..", "..", "bin", runtime.GOOS+"_"+runtime.GOARCH, "plugincheck2")
	env := []string{
		"DEBUG=0",
		"OPENAI_API_KEY=",
	}

	// ...other setup...

	// Run each test case
	for file := range files {
		t.Run(file, func(t *testing.T) {
			// Allows the test case to run in parallel with other ones
			t.Parallel()

			// (use t.Log instead of fmt.Print)
			t.Logf("Running %s", file)
		
			// Testing code that runs "sh -c plugincheck2 -config ......"
			// (use `require` and `assert`)
			err := cmd.Run()
			require.NoError(t, err)
			require.NotEmpty(t, outb)

			// ...rest of the test case
		}
	}
}
```

Finally, you also have to update the `Magefile` so it runs `go test -v ./pkg/cmd/plugincheck2` instead of calling `integrationtests.RunTests`:

```go
func (Test) Integration() error {
	mg.Deps(Build.Local)
	return sh.RunV("go", "test", "-v", "./pkg/cmd/plugincheck2")
}
```

pkg/integrationtests/integrationtests.go
```diff
@@ -0,0 +1,233 @@
...
+			filepath.Join(basePath, file),
+		)
+		fmt.Printf("Running command: %s\n", command)
+		cmd := exec.Command("sh", "-c", command)
+		var outb, errb bytes.Buffer
+		cmd.Stdout = &outb
+		cmd.Stderr = &errb
+		cmd.Env = append(os.Environ(), env...)
+		err := cmd.Run()
+		if err != nil && len(outb.String()) == 0 {
```
xnyo, line 193: Nit: I would probably also assert on the exit code of the validator. For the `invalid` zips in particular, the exit code is 1.

To do this, I would add a new `TestCase` struct:

```go
type TestCase struct {
	expExitCode int
	expReport   JsonReport
}
```

and change the `files` map so it's a `map[string]TestCase{`

and for the invaid ones:

```go
	"invalid.zip": {
		expExitCode: 1,
		expReport: JsonReport{
			Id:      "invalid-panel",
			Version: "1.0.0",
			PluginValidator: map[string][]Issue{
				"archive": {
					{
						Severity: "error",
						Title:    "Archive contains more than one directory",
						Detail:   "Archive should contain only one directory named after plugin id. Found 7 directories",
						Name:     "more-than-one-dir",
					},
					{
						Severity: "error",
						Title:    "Plugin archive is improperly structured",
						Detail:   "",
						Name:     "zip-invalid",
					},
				},
			},
		},
	},
```

Then, to assert it:

```go
err := cmd.Run()
if tc.expExitCode == 0 {
	require.NoError(t, err)
} else {
	require.Equal(t, tc.expExitCode, cmd.ProcessState.ExitCode())
	require.Error(t, err)
}
require.NotEmpty(t, outb)
```

pkg/integrationtests/integrationtests.go
```diff
@@ -0,0 +1,233 @@
...
+			return fmt.Errorf(
+				"Error running integration tests for file %s: %s",
+				file,
+				errb.String(),
+			)
+		}
+
+		// marshall the output into a JsonReport
+		var report JsonReport
+		err = json.Unmarshal(outb.Bytes(), &report)
```
xnyo, line 203: This can be simplified as:

```go
var report JsonReport
require.NoError(t, json.NewDecoder(&outb).Decode(&report))
```

xnyo (commented):
Amazing work! :rocket:  This is looking really good. I have some suggestions but the main part is declaring the integration tests directly as Go tests, so we can use `go test` and you don't have to figure out the report :) LMK what you think!

pkg/integrationtests/integrationtests.go
academo: Thanks for such a nice break down. I will update this PR

Magefile.go
xnyo, line 1: I had another look at the `Magefile` and now the tests will run also when running simply `go test`, which means they are also part of the  `mage test:default` and `mage test:verbose` targets.

However, integration tests require the binary to be present on disk, which is build by the `Build.Local` target.

We have to change the `test:default` and `test:verbose` targets so that the `Build.Local` target is executed before `go test`, otherwise the tests will fail if the binary is no present:

```diff
diff --git a/Magefile.go b/Magefile.go
index b134e98..bce2a52 100644
--- a/Magefile.go
+++ b/Magefile.go
@@ -214,14 +214,16 @@ func (Build) Lint() error {
 
 // Run tests in verbose mode
 func (Test) Verbose() {
-       mg.Deps(
+       mg.SerialDeps(
+               Build.Local,
                testVerbose,
        )
 }
 
 // Run tests in normal mode
 func (Test) Default() {
-       mg.Deps(
+       mg.SerialDeps(
+               Build.Local,
                test,
        )
 }
```

`mg.SerialDeps` is required, otherwise `Build.Local` and `test` will run in parallel: we need `Build.Local` to finish before `test` or the integration tests may fail if the binary hasn't finished building yet.

Another option is to use `testing.Short` to flag integration tests, and run those only with `mage test:integration`:

```go
func TestIntegration(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping integration test")
	}
	// ... the test ...
}
```

Then, you can use:

```
go test -short -v ./pkg/...
```

To run the "unit tests", and this one:

```bash
go test -v ./pkg/cmd/plugincheck2
```

To run only the integration tests

xnyo (commented):
Amazing work refactoring this! :clap::clap::clap:

I have a few small comments but it's looking good so far!

Magefile.go
academo: I prefer we make these part of the regular full test suite so I'll keep the serial dependency as you suggested.

xnyo (approved):
LGTM! :rocket:

PR #197, opened 2024-05-08
Enhance nested plugins support (part 2)

academo:
Follow up to https://github.com/grafana/plugin-validator/pull/194 

If https://github.com/grafana/plugin-validator/pull/194 is not merged yet please don't review this yet.

Added nested plugins support for:

* Binary permissions validator
* Backend debug validator
* Backend binaries files validator
* SDK Usage validator


pkg/analysis/passes/backendbinary/backendbinary.go
```diff
@@ -43,90 +42,88 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
-	if data.Executable != "" && !data.Backend {
-		pass.ReportResult(
-			pass.AnalyzerName,
-			backendFoundButNotDeclared,
-			"Found executable in plugin.json but backend=false",
-			"You have marked your plugin with backend=false in your plugin.json but declared a backend executable. Please set backend=true in your plugin.json if your plugin has a backend component or remove the executable from your plugin.json",
-		)
-		return nil, nil
-	}
+	for pluginJsonPath, data := range metadatamap {
```
academo, line 51: though the changes look big. this is merely moving all the previous logic into a loop of the metadatamap instead of a single metadata

pkg/analysis/passes/backenddebug/backenddebug.go
```diff
@@ -28,35 +35,49 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
-	for _, fn := range []string{"standalone.txt", "pid.txt"} {
-		if _, err := os.Stat(filepath.Join(archiveDir, fn)); err != nil {
-			if os.IsNotExist(err) {
-				// Banned file not found
-				continue
-			}
-			return nil, fmt.Errorf("stat %q: %w", fn, err)
-		}
-		// Found a banned file
+	textFiles, err := doublestar.FilepathGlob(archiveDir + "/**/*.txt")
```
academo, line 57: changed the logic to search for the txt files across all plugins inside an archive.

xnyo (commented):
Great work! I have some comments but looks good so far! :raised_hands:

pkg/analysis/passes/backendbinary/backendbinary_test.go
```diff
@@ -6,48 +6,60 @@ import (
...
 	var interceptor testpassinterceptor.TestPassInterceptor
-	const pluginJsonContent = `{
-    "name": "my plugin name"
-  }`
+	pluginJsonContent := []byte(`{
+		"id": "test-plugin-panel",
+		"type": "panel"
+  }`)
+
+	meta, err := utils.JsonToMetadata(pluginJsonContent)
```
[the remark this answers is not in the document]
academo: thanks for noticing. it was a merge mistake on my part.

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -40,67 +46,110 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
-				binaryExecutablePermissions,
-				fmt.Sprintf("Could not read permissions for executable %s", binaryPath),
-				fmt.Sprintf("Could not read the file %s. This could be too few permissions in your binary files or your zip file is corrupted", binaryPath))
-			continue
+			pass.ReportResult(
+				pass.AnalyzerName,
+				archiveFilesError,
+				"error reading your archive files",
+				fmt.Sprintf("error reading your archive files: %s", err.Error()),
+			)
```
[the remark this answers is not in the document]
academo: I understand that it sounds like this should be an internal validator error but not being able to traverse the errors inside the archive can signal that the archive itself is wrong. e.g. corrupted, wrong permissions, etc.. not the very rare runner error.

My intention with the error is to be able to send this as feedback to the person running the validator and in our submission pipeline.

An internal error in the validator stops the whole validation process and disrupts the pipeline without providing useful information to the final user in the submission process.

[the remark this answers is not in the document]
xnyo: I see, good point. Sounds good, but in that case I think you probably need a `continue` after this `pass.ReportResult` call in order to skip to the next `pluginJsonPath`

pkg/analysis/passes/binarypermissions/binarypermissions.go
```diff
@@ -40,67 +46,110 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+				binaryExecutableFound,
+				fmt.Sprintf(
+					"No binary found for `executable` %s defined in %s", executable, pluginJsonPath,
+				),
+				fmt.Sprintf(
+					"You defined an executable %s but it could not be found for any of the supported architectures",
+					executable,
+				),
+			)
+			return nil, nil
```
xnyo, line 111: Also here I think this should be a `continue` instead?

academo: Yes! thanks for catching these. Added.

xnyo (approved):
LGTM! :rocket:

PR #198, opened 2024-05-13
Validate SDK version in sdk usage validator

### LINKED ISSUE #126: New analyzer: detect usage of go-sdk and version for backend plugins
Provide a new analyzer to detect whether the go-sdk is being used for a backend component and which version is being used. 

It _could_ check for `backend=true` in the `plugin.json` file. If yes, scan for a go.mod file in the project and check the version declared there. 

To start with, it should warn if the sdk is not being used and if the version is not latest (as well as printing "we recommended updating to the current version <version>".

In the future we can consider failing if the sdk is not being used and if the version is significantly out of date

academo:
This PR introduces validation of the SDK version that plugins with a backend component use.

The latest SDK version of the grafana sdk is grabbed using the github API https://api.github.com/repos/grafana/grafana-plugin-sdk-go/releases/latest

* Plugins using the latest grafana go sdk version or released in the last 2 months will pass
* Plugins using a version between 2 and 5 months old will get a warning
* Plugins using a version older than 5 months will get an error

Closes https://github.com/grafana/plugin-validator/issues/126


pkg/analysis/passes/sdkusage/sdkusage.go
```diff
@@ -101,5 +125,80 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	if err != nil {
+		// it is most likely this failed because of github auth or rate limits
+		logme.Debugln(err)
+		return nil, nil
+	}
+
+	daysDiff, err := daysDifference(pluginGoSdkRelease.PublishedAt, latestRelease.PublishedAt)
+	if err != nil {
+		// error calculating the days difference could be a problem in github date format
+		// ignoring it
```
andresmgot, line 160: we should log the error, right?

pkg/analysis/passes/sdkusage/sdkusage.go
```diff
@@ -3,23 +3,44 @@ package sdkusage
...
+		Name:     "go-sdk-older-than-2-months",
+		Severity: analysis.Warning,
+	}
+	goSdkOlderThanFiveMonths = &analysis.Rule{
+		Name:     "go-sdk-older-than-5-months",
+		Severity: analysis.Error,
+	}
 )
 
+var twoMonths = 30 * 2
```
andresmgot, line 30: you could work with `time.Duration` here if you want to:

```go
var day, _ = time.ParseDuration("24h")
var twoMonths = day * 30 * 2
```

Not necessarily better since this is simple enough

pkg/analysis/passes/sdkusage/sdkusage_test.go
```diff
@@ -76,6 +77,29 @@ func TestGoModNotParseable(t *testing.T) {
...
 
 func TestValidGoMod(t *testing.T) {
+	httpmock.Activate()
+	defer httpmock.DeactivateAndReset()
+
+	// mock latest request
+	httpmock.RegisterResponder(
+		"GET",
+		"https://api.github.com/repos/grafana/grafana-plugin-sdk-go/releases/latest",
+		httpmock.NewStringResponder(
```
andresmgot, line 87: I'm not used to this `httpmock` package but it seems that you have a `httpmock.NewJsonResponder` which seems a better fit? It sets the proper headers.

pkg/analysis/passes/sdkusage/sdkusage.go
```diff
@@ -101,5 +125,80 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	}
+
+	pluginGoSdkRelease, err := githubutils.FetchGrafanaSdkReleaseByTag(pluginGoSdkVersion)
+	if err != nil {
+		// it is most likely this failed because of github auth or rate limits
+		logme.Debugln(err)
+		return nil, nil
+	}
+
+	daysDiff, err := daysDifference(pluginGoSdkRelease.PublishedAt, latestRelease.PublishedAt)
```
andresmgot, line 157: I think it's good to use as origing the release date of the latest release but maybe we should specify that somewhere? even if it's just a comment here.

This has a caveat though, if we don't release a version for a while (> 2 months) and a plugin is using an older version (e.g 3 months), it won't receive a warning. If the next day we release a version, it will start failing (which can be confusing if the plugin is in the review process). I wonder if it's better to just use the current day for the comparison here.

pkg/analysis/passes/sdkusage/sdkusage_test.go
andresmgot: For the record, we usually use the `httptest` package for this kind of tests, which is an upstream package and gives you more flexibility (it's also more low level though).

pkg/analysis/passes/sdkusage/testdata/sdk-2-months-old/go.mod
andresmgot, line 1: nit: I would simplify these test files so it's clear what's in use here.

pkg/githubutils/githubutils.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+
+	"github.com/grafana/plugin-validator/pkg/logme"
+)
+
+func FetchLatestGrafanaSdkRelease() (Release, error) {
+	return FetchGrafanaSdkReleaseByTag("latest")
+}
+
+func FetchGrafanaSdkReleaseByTag(tag string) (Release, error) {
```
andresmgot, line 16: I now this is fairly simple at the moment but have you considered using a library for this? like https://github.com/google/go-github. It should help handling weird cases.

andresmgot: Also, I don't know if we are able to use any persistence with the validator but it may be worth to cache these results so we don't need to query them again (not a big problem if we use a token though).

pkg/githubutils/types.go
```diff
@@ -0,0 +1,44 @@
+package githubutils
+
+type Author struct {
```
andresmgot, line 3: yeah, you shouldn't need to define these if you use a library.

andresmgot (commented):
Nice! I have some comments / questions

pkg/analysis/passes/sdkusage/sdkusage_test.go
academo: Good idea about using the jsonresponser but I think it requires me to pass it an actual struct and not a json string which (IMO) is closer to what the network actually returns. I'lldo the change though, it makes it easier to read.

Regarding using `httptest` I tried using it but the amount of boilerplate and setup was not worth it for this simple test setup. Should we need some of those lower level APIs I'm sure we would use it. We also already use the httpmock package in other tests (of course introduced by me)

pkg/analysis/passes/sdkusage/sdkusage.go
academo: I'll see if maybe this allows "2 months" as input otherwise I'll leave it as is.

pkg/githubutils/githubutils.go
academo: I was planning to use the go-github but, yes, this is such a simple http request that using a whole library makes for more work and yet another dependency.

If we start adding more checks that rely on the github API we should move to it though.

pkg/githubutils/types.go
academo: Indeed. But I think for now it is an acceptable tradeoff considering the simple use case.

pkg/analysis/passes/sdkusage/testdata/sdk-2-months-old/go.mod
academo: Do you mean remove all the other non-relevant require lines leaving only the grafana go sdk one?

andresmgot: yep

pkg/analysis/passes/sdkusage/sdkusage.go
academo: Good idea. I'll change the logic to use the current date as the comparison date.

pkg/githubutils/githubutils.go
```diff
@@ -0,0 +1,74 @@
+package githubutils
```
xnyo, line 1: I would rename this package, in general in Go it's discouraged to use names such as `utils` as they can be confusing. I would suggest either calling the package just `github`, or maybe or `githubapi`

pkg/githubutils/githubutils.go
```diff
@@ -0,0 +1,74 @@
...
+func FetchLatestGrafanaSdkRelease() (Release, error) {
+	return FetchGrafanaSdkReleaseByTag("latest")
+}
+
+func FetchGrafanaSdkReleaseByTag(tag string) (Release, error) {
+	if tag == "" {
+		return Release{}, fmt.Errorf("tag is required")
+	}
+
+	apiUrl := "https://api.github.com/repos/grafana/grafana-plugin-sdk-go/releases"
```
xnyo, line 21: Nit: This can be a `const`:
```suggestion
	const apiUrl = "https://api.github.com/repos/grafana/grafana-plugin-sdk-go/releases"
```

pkg/githubutils/githubutils.go
```diff
@@ -0,0 +1,74 @@
...
+	} else {
+		apiUrl = apiUrl + "/tags/" + tag
+	}
+
+	req, err := http.NewRequest("GET", apiUrl, nil)
+	if err != nil {
+		return Release{}, err
+	}
+
+	token := os.Getenv("GITHUB_TOKEN")
```
xnyo, line 34: Suggestion: This is probably okay for now but I would use a:

```go
type Client struct {
    token string
}

func NewClient(token string) *Client {
    return &Client{token: token}
}

func (c *Client) FetchGrafanaSDKReleaseByTag(...) {
    // ...
    // use c.token
    //
}
```

```go
// somewhere in main..
ghClient := github.Client(os.Getenv("GITHUB_TOKEN"))
```

and pass the client around. In general it's discouraged to use environment variables in this global-like way. But probably okay for now, I would also consider switching to a package instead in the future like Andres suggested in case we end up making more complicated things with the GitHub API.

xnyo (commented):
Awesome work! I have some minor suggestions but looks great otherwise!

pkg/githubutils/githubutils.go
academo: good idea. I really want to keep this simple without the need for more and more parameters because if I make this function accept the token I'll need to pass it all the way down from the main function to this one and it'll just make the changes downstream too big

academo:
@andresmgot @xnyo thanks for the reviews.

I incorporated most of your feedback leaving behind the usage of the github go library which is an overkill for the simple usage we have for it.

Should we use more apis from github in the future I'll swap it for the github go library.

andresmgot (approved):
I'd still use the github library to avoid having to define the types and the error handling but up to you. For the rest, LGTM

pkg/analysis/passes/sdkusage/sdkusage.go
```diff
@@ -101,5 +125,82 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		logme.DebugFln("Error calculating days difference: %s", err.Error())
+		return nil, nil
+	}
+
+	if daysDiff > fiveMonths {
+		pass.ReportResult(
+			pass.AnalyzerName,
+			goSdkOlderThanFiveMonths,
+			"Your Grafana go sdk is older than 5 months",
+			"Your Grafana go sdk is older than 5 months. Please upgrade to the latest version",
```
xnyo, line 166: Nit: Maybe we should change those messages with guidance on how to upgrade the package? Or at least give the full name of the Go package.

Something like:

```
Your Grafana Go SDK version is older than 5 months. Please upgrade to the latest version by running: "go get -u github.com/grafana/grafana-plugin-sdk-go"
```

Or:

```
Your Go package "github.com/grafana/grafana-plugin-sdk-go" is older than 5 months. Please upgrade to the latest version.
```

I would also change the way it's spelled

`Grafana go sdk` -> `Grafana Go SDK`

WDYT?

xnyo (approved):
Awesome work! Just one suggestion regarding the message, but LGTM otherwise! :raised_hands:

pkg/analysis/passes/sdkusage/sdkusage.go
academo: Good idea. let me change that.

PR #201, opened 2024-05-21
Add LLMValidate and remove tzap gpt 

academo:
* Removes TZAP as a validator that was using OpenAI GPT to analyze the code
* Adds a slighly more generic (but not fully generic) validator that currently uses Gemini Flash

The new validator is using a new module that attempts to be more generic in case we want to start implementing later other LLMs APIs. For now this new module is tied to gemini but it has some ground work to make it generic should we want to in the future


pkg/analysis/passes/llmreview/llmreview.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if err != nil {
+		logme.DebugFln("Error getting answers from Gemini LLM: %v", err)
+		return nil, nil
+	}
+
+	for _, answer := range answers {
+		shortAnswer := strings.TrimSpace(strings.ToLower(answer.ShortAnswer))
+		if shortAnswer == "yes" || shortAnswer == "na" {
```
xnyo, line 77: Maybe it's better to change the check so that:

```suggestion
		if shortAnswer != "no" {
```

Not sure if it may answer with something completely different than `yes`, `no` or `na`

pkg/llmvalidate/llmvalidate.go
```diff
@@ -0,0 +1,307 @@
...
+	"**/__mocks__/**",
+	"**.test.*",
+	"test-**",
+	"jest.config.*",
+	"babel.config.*",
+	"jest-setup.*",
+	"playwright.config.*",
+	"vite.config.*",
+	"tsconfig.*",
+	"package-lock.json",
```
xnyo, line 36: Is this required? `.json` is not in the `allowedExtensions` list anyways :thinking:

pkg/llmvalidate/llmvalidate.go
```diff
@@ -0,0 +1,307 @@
...
+		return ""
+	}
+
+	if isMinifiedJsFile(content) {
+		return ""
+	}
+
+	if len(content) == 0 {
+		return ""
+	}
```
xnyo, line 262: Maybe we also have to add a max size limit, so if the file is too big it doesn't get sent to the LLM? We would then have to bubble-up this kind of warning to the output even if it's not sent to the LLM, so it may be a bit complicated. I would leave it for a follow-up PR eventually, but interested in hearing if you think this can be a concern

pkg/llmvalidate/llmvalidate.go
```diff
@@ -0,0 +1,307 @@
...
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	var content strings.Builder
+
+	for scanner.Scan() {
+		if !utf8.ValidString(string(scanner.Text())) {
+			return "", fmt.Errorf("invalid UTF-8 in file %s", filePath)
+		}
+		content.WriteString(scanner.Text() + "\n")
```
xnyo, line 299: A couple of things here:
- For efficiency, `scanner.Text()` can be put in a variable so the string isn't allocated twice
- You don't need to cast `scanner.Text()`'s return value to a `string`, because it already returns a `string`
- Rather than writing `scanner.Text() + "\n"` to the `strings.Builder` (which allocates a new string passed to `WriteString`), you can write the string first, and then write `\n` with a separate `WriteRune` call

```suggestion
		text := scanner.Text()
		if !utf8.ValidString(text) {
			return "", fmt.Errorf("invalid UTF-8 in file %s", filePath)
		}
		content.WriteString(text)
		content.WriteRune('\n')
```

xnyo (commented):
Awesome work!! :rocket: 

I have left some comments but looks very promising! :clap: :clap:

pkg/llmvalidate/llmvalidate.go
academo: Ah! left over from a previous implementation. THanks for catching it.

pkg/llmvalidate/llmvalidate.go
academo: THe great thing about gemini is the huge context size. currently above 2M. I am sure we'll eventually hit the limit with some incredibly big plugin but it will work for most cases.

in those cases the LLM will fail with a bad request and after some retries the validator will give up,

xnyo (approved):
LGTM! Awesome work! :raised_hands:

PR #203, opened 2024-05-28
llmvalidate: Add more files to exclude list

academo:


pkg/llmvalidate/llmvalidate.go
```diff
@@ -274,6 +279,8 @@ func getPromptContentForFile(codePath, relFile string) string {
 		return ""
 	}
 
+	logme.DebugFln("Including file %s", relFile)
```
academo: This log is intentional

marefr (approved):

xnyo (approved):
LGTM!

PR #213, opened 2024-06-11
LLMReview: only send files inside src and pkg for llm to review

academo:
- **LLM Validate: Allow to specify subpaths for llmvalidate**
- **llmreview: only validate files inside src and pkg**

### DIFF
--- a/pkg/llmvalidate/llmvalidate_test.go
+++ b/pkg/llmvalidate/llmvalidate_test.go
@@ -1,0 +1,36 @@
+package llmvalidate
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetPromptsForCodepathWithSubPaths(t *testing.T) {
+	codePath := filepath.Join("testdata", "testplugin")
+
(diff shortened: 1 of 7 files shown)


pkg/llmvalidate/llmvalidate_test.go
```diff
@@ -0,0 +1,36 @@
...
+	prompts, err := getPromptContentForCode(codePath, []string{"src", "pkg"})
+
+	require.NoError(t, err)
+	require.Len(t, prompts, 2)
+}
+
+func TestGetPromptsForCodepathWithoutSubPaths(t *testing.T) {
+	codePath := filepath.Join("testdata", "testplugin")
+
+	prompts, err := getPromptContentForCode(codePath, []string{"."})
```
wbrowne, line 22: Nit: Maybe a test for `prompts, err := getPromptContentForCode(codePath, []string{})` since it will automatically set this if empty right?

wbrowne (approved):
Looks good - just a nit on a test.

pkg/llmvalidate/llmvalidate_test.go
academo: Good idea. let me add it.

PR #219, opened 2024-06-18
Fix source code validation with windows line ending

### LINKED ISSUE #205: Source code validation: Fails on different line endings
image

academo:
Closes https://github.com/grafana/plugin-validator/issues/205

Fixes an issue where the validator will fail to compare files that were saved in windows against files that were saved on linux.

This occurs when the archive was created with line endings that were different on linux and windows.

There's no risk on having a false positive or no detecting changes because js files are normalized to unix line endings or windows line endings before comparison
but the content stays the same.

### DIFF
--- a/pkg/analysis/passes/gomanifest/gomanifest.go
+++ b/pkg/analysis/passes/gomanifest/gomanifest.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -69,18 +68,24 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	logme.DebugFln("manifestFilePath: %s", manifestFilePath)
 	maniFestFiles, err := parseManifestFile(manifestFilePath)
 	if err != nil {
-		pass.ReportResult(pass.AnalyzerName, noGoManifest,
+		pass.ReportResult(
+			pass.AnalyzerName,
+			noGoManifest,
 			"Could not find or parse Go manifest file",
-			"Your source code contains Go files but there's no Go build manifest. Make sure you are using the latest version of the Go plugin SDK")
+			"Your source code contains Go files but there's no Go build manifest. Make sure you are using the latest version of the Go plugin SDK",
+		)
 		return nil, nil
 	}
 
 	err = verifyManifest(maniFestFiles, goFiles, sourceCodeDir)
 	if err != nil {
 		logme.DebugFln("verifyManifest error: %s", err)
-		pass.ReportResult(pass.AnalyzerName, invalidGoManifest,
+		pass.ReportResult(
+			pass.AnalyzerName,
+			invalidGoManifest,
 			"The Go build manifest does not match the source code",
-			"The provided Go build manifest does not match the provided source code. If you are providing a git repository URL make sure to include the correct ref (branch or tag) in the URL and it includes all the Go files used to build the plugin binaries")
+			"The provided Go build manifest does not match the provided source code. If you are providing a git repository URL make sure to include the correct ref (branch or tag) in the URL and it includes all the Go files used to build the plugin binaries",
+		)
 		return nil, nil
 	}
 
@@ -141,7 +146,7 @@ func verifyManifest(manifest map[string]string, goFiles []string, sourceCodeDir 
 			return err
 		}
 		// calculate the sha256sum of the go file
-		sha256sum, err := hashFileContent(goFilePath)
+		sha256sum, windowsSha256Sum, err := hashFileContent(goFilePath)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


pkg/analysis/passes/gomanifest/gomanifest.go
```diff
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
```
academo, line 7: notice I applied long line formatting to this file

pkg/analysis/passes/gomanifest/gomanifest.go
```diff
@@ -141,18 +146,22 @@ func verifyManifest(manifest map[string]string, goFiles []string, sourceCodeDir
...
-			return fmt.Errorf("could not find file %s with hash %s in manifest", goFileRelativePath, sha256sum)
+			return fmt.Errorf(
+				"could not find file %s with hash %s in manifest",
+				goFileRelativePath,
+				sha256sum,
+			)
 		}
 		// check if the sha256sum in the manifest matches the calculated sha256sum
-		if sha256sum != manifestSha256sum {
+		if sha256sum != manifestSha256sum && windowsSha256Sum != manifestSha256sum {
```
academo, line 164: this is the fundamental change in the validation of the go manifest. now we calculate two hashes, one with linux line endings and other with windows line endings. We accept either as a valid hash.

pkg/analysis/passes/gomanifest/gomanifest.go
```diff
@@ -167,43 +176,23 @@ func verifyManifest(manifest map[string]string, goFiles []string, sourceCodeDir
...
 
-func hashFileContent(path string) (string, error) {
-	// Handle hashing big files.
-	// Source: https://stackoverflow.com/q/60328216/1722542
-
-	f, err := os.Open(path)
+// we hash the file content using the sha256 algorithm
+// a second hash is created using the line endings that windows uses
+// for plugins compiled with native windows that use windows line endings
+func hashFileContent(path string) (string, string, error) {
```
academo, line 182: because I now have to essentially run replace in the whole content I went with reading the whole file instead of chunks. It is very unlikely we'll have go files big enough to overflow the memory

pkg/difftool/difftool.go
```diff
@@ -65,7 +70,19 @@ func CompareSourceMapToSourceCode(pluginID string, sourceCodeMapFile string, sou
 			continue
 		}
 
-		delta := godiffpatch.GeneratePatch(sourceMapFileName, string(sourceCodeFileContent), sourceMapContent)
+		// Normalize line endings to Unix-style (LF) to ensure consistent comparison
+		cleanSourceCodeFileContent := strings.ReplaceAll(
```
academo, line 74: I am normalizing the line ending here before calculating the hash.

pkg/analysis/passes/gomanifest/gomanifest.go
marefr: Should it be or instead of and?

pkg/analysis/passes/gomanifest/gomanifest.go
```diff
@@ -167,43 +176,23 @@ func verifyManifest(manifest map[string]string, goFiles []string, sourceCodeDir
...
 
-	defer func() {
-		err = f.Close()
-		if err != nil {
-			fmt.Printf("error closing file for hashing: %v", err)
-		}
-	}()
+	// Calculate the original hash
+	hOrig := sha256.Sum256(data)
+	originalHash := hex.EncodeToString(hOrig[:])
```
marefr, line 190: Did you consider normalize here as well similar to how you do source map comparison?

marefr (changes requested):
Works/good enough. Added a question. Would be good with some tests as well

pkg/analysis/passes/gomanifest/gomanifest.go
academo: the go manifest is a list of the hashed content of the go files at build, the hash is calculated using the line ending of the OS that wrote and we do here is rehash the file and verify. Here we are basically detecting which line ending was used by calculating both hashes and checking.

pkg/analysis/passes/gomanifest/gomanifest.go
academo: This is the condition to fail so passing one is a success (hence and `and`). If we would put an `or` here it'll always fail:

A file is hashed and written into the manifest. the hash is calculated based on the content that might be windows line end or linux line end.

We calculate both hashes of the original file. one with linux line ending and other with windows line ending.

if neither of those hashes matches the hash in the manifest we fail the check.

academo:
> Works/good enough. Added a question. Would be good with some tests as well

@marefr I improved the code variables and legibility to make the intention more clear and added a test case for the windows line ending.

marefr (approved):
LGTM

PR #236, opened 2024-07-22
Trackingscripts: Check for urls

academo:
Improve the logic for the tracking scripts to check only inside urls and not all the code to prevent false positives.

### DIFF
--- a/pkg/analysis/passes/trackingscripts/trackingscripts_test.go
+++ b/pkg/analysis/passes/trackingscripts/trackingscripts_test.go
@@ -44,6 +44,14 @@ func TestTrackingScriptsInvalid(t *testing.T) {
 	_, err := Analyzer.Run(pass)
 	require.NoError(t, err)
 	require.Len(t, interceptor.Diagnostics, 1)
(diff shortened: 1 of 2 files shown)


pkg/analysis/passes/trackingscripts/trackingscripts_test.go
```diff
@@ -44,6 +44,14 @@ func TestTrackingScriptsInvalid(t *testing.T) {
...
+	require.Equal(
+		t,
+		interceptor.Diagnostics[0].Title,
+		"module.js: should not include tracking scripts",
+	)
+	require.Equal(
+		t,
+		interceptor.Diagnostics[0].Detail,
+		"Tracking scripts are not allowed in Grafana plugins (e.g. google analytics). Please remove any usage of tracking code. Found: https://www.google-analytics.com/analytics.js",
+	)
```
leventebalogh, line 56: Would be nice to add a test case that checks for this improvement (e.g. that `grafana-asserts-app.rules:read` doesn't trigger `"pp.ru"`).

leventebalogh (approved):
🚀

jackw (approved):

PR #237, opened 2024-07-22
Tracking scripts: Add missing test case

academo:
- **Trackingscripts: Check for urls**
- **Do not stop after finding the first**
- **Add missing test case**


leventebalogh (approved):
🥇

PR #248, opened 2024-08-28
Chore: remove binwrap dependency

academo:
Removes binwrap as a node dependency.

We used to use binwrap to distribute the binary to use it with npx but
it seems binwrap is not maintainted anymore and security issues had been
reported due to sub-dependencies vulnerabilities.

Tihs PR replicates the functionality of binwrap and reduces the node
dependencies to only `tar`


bin/index.js
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env node
```
academo, line 1: This script effectively replaces all the work binWrap was doing

tolzhabayev (approved):
did not test but looks plausible

xnyo (approved):
LGTM! I have tested it and it works well, awesome work! :rocket:

jackw (approved):
LGTM. 🚀

PR #263, opened 2024-10-10
Semgrep: only check for console log usage inside the src folder

### LINKED ISSUE #262: Limit console.log check to src folder
Multiple times, this check returns errors for files that are not part of the plugin (like test files or additional tooling). To avoid the noise, we can check the `src` folder only, which is the only folder used to compile plugins by default.

academo:
Closes https://github.com/grafana/plugin-validator/issues/262


pkg/analysis/passes/virusscan/virusscan_test.go
```diff
@@ -18,6 +18,9 @@ func isClamAvInstalled() bool {
 	if err != nil {
 		return false
 	}
+	if os.Getenv("SKIP_CLAMAV") != "" {
```
academo, line 21: intentional change.

running tests when clamav is locally installed takes a considerable amount of time, this makes it much faster to run all tests locally except for these very specific ones.

andresmgot: having clamav does not sound nice 😆

andresmgot (approved):
LGTM! Thanks!

PR #264, opened 2024-10-29
Feat: Add flag to select specific analyzer to run

oshirohugo:
This PR adds a new flag to `plugincheck2` command line: `analyzer`. It will allow us to select individuals analyzers to run. 

This will allow us to use `plugincheck2` in the CI provided by `create-plugin` to validade the `plugin.json` format as required here: https://github.com/grafana/plugin-tools/issues/1062

OBS: The way that our analyzers are implemented will guarantee that all the required analyzer will run even when individual analyzers are selected to run with the new flag.

### DIFF
--- a/pkg/cmd/plugincheck2/main_test.go
+++ b/pkg/cmd/plugincheck2/main_test.go
@@ -28,6 +28,12 @@ type JsonReport struct {
 	PluginValidator map[string][]Issue `json:"plugin-validator"`
 }
 
+type TC struct {
+	File       string
+	ExtraArgs  string
+	JsonReport JsonReport
+}
+
 func TestIntegration(t *testing.T) {
 	basePath := "./testdata"
 	binary := filepath.Join(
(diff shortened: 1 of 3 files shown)


academo (commented):
nice! please make sure to add an integration test with this case here 

https://github.com/grafana/plugin-validator/blob/main/pkg/cmd/plugincheck2/main_test.go

You can use the same fixtures 

Also remember to update the README file to include this flag in the documentation

academo (approved):

pkg/cmd/plugincheck2/main_test.go
```diff
@@ -28,6 +28,12 @@ type JsonReport struct {
 	PluginValidator map[string][]Issue `json:"plugin-validator"`
 }
 
+type TC struct {
+	File       string
+	ExtraArgs  string
+	JsonReport JsonReport
+}
```
xnyo, line 35: Nit: can we unexport this?

```suggestion
type tc struct {
	file       string
	extraArgs  string
	jsonReport JsonReport
}
```

xnyo (approved):
Awesome work! :tada: Just one minor nit, but LGTM otherwise

PR #272, opened 2024-11-07
Add circular dependencies analyzer

### LINKED ISSUE #270: Prevent plugin dependency cycles
We want to avoid situations where the plugin dependency feature is misused

We should prevent scenarios where:
1. Plugin A has a dependency on itself
2. Plugin A has a dependency B and Plugin B has a dependency on A

For example:

## 1

```json
{
  "id": "grafana-foo-datasource",
  "dependencies": {
    "grafanaDependency": ">=11.1.0",
    "plugins": [
      {
        "id": "grafana-foo-datasource",
      }
    ]
  }
}
```

## 2

```json
{
  "id": "grafana-foo-datasource",
  "dependencies": {
    "grafanaDependency": ">=11.1.0",
    "plugins": [
      {
        "id": "grafana-bar-datasource",
      }
    ]
  }
}
```

```json
{
  "id": "grafana-bar-datasource",
  "dependencies": {
    "grafanaDependency": ">=11.1.0",
    "plugins": [
      {
        "id": "grafana-foo-datasource",
      }
    ]
  }
}
```

xnyo:
Adds an analyzer to validate that plugins do not have circular dependencies. Fixes #270.


pkg/analysis/passes/circulardependencies/circulardependencies.go
```diff
@@ -0,0 +1,121 @@
...
+			} else {
+				// The dependency is not a nested plugin, get the dependencies from GCOM
+				var err error
+				version := dep.Version
+				if version == "" {
+					version = "latest"
+				}
+				dependantDependencies, err = getGCOMPluginDependencies(ctx, dep.ID, version)
+				if err != nil {
+					return nil, fmt.Errorf("get plugin version dependency from gcom: id=%q version=%q: %w", dep.ID, version, err)
```
academo, line 73: returning an error here effectively stops the validator from running and not running any other validator. Are we sure we want to do that or instead should we report the error (a new rule type?) and `continue`?

academo (commented):
great work! a small comment to consider not returning an error in the specific gcom case.

academo (approved):

PR #273, opened 2024-11-07
Refactoring and cleanup

xnyo:
Several style changes to adhere to Go best practices and to suppress some warnings/code smells.


pkg/analysis/passes/manifest/manifest.go
```diff
@@ -215,11 +215,11 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return b, nil
 }
 
-// parseManifestFile parses the manifest file and returns a ManifestFile struct
+// parseManifestFile parses the manifest file and returns a File struct
 // it does not verify the signature only returns the content
-func parseManifestFile(b []byte) (ManifestFile, error) {
+func parseManifestFile(b []byte) (File, error) {
```
academo, line 220: IMO this should remain called "ManifestFile"

pkg/analysis/passes/signature/signature.go
```diff
@@ -61,11 +61,14 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 		} `json:"info"`
 	}
-	if err := json.Unmarshal(metadata, &data); err != nil {
+	if err := json.Unmarshal(md, &data); err != nil {
 		return nil, nil
 	}
 
-	state := getPluginSignatureState(data.ID, data.Info.Version, archiveDir, manifest)
+	state, err := getPluginSignatureState(data.ID, data.Info.Version, archiveDir, mf)
+	if err != nil {
```
academo, line 69: we don't want to return an error here or this will stop all the validator from running. You can add a report and return nil,nil

academo (commented):
Nice work! 

only a real concern in the new return error that can potentially stop the validator from running when in the past it'd normally run.

pkg/analysis/passes/signature/signature.go
s4kh: cc @xnyo, everything else looks good.

xnyo: Great catch, I always forget about that! Fixed

pkg/analysis/passes/manifest/manifest.go
xnyo: The package is already called `manifest`, so we should avoid repeating it: https://go.dev/blog/package-names#naming-package-contents

I'd like to keep it like this

```go
manifest.ManifestFile
```

:point_down: 

```go
manifest.File
```

s4kh (approved):
Superb refactoring. :100:

academo (approved):
great work on this!

PR #274, opened 2024-11-07
Add severity flag

oshirohugo:
Add `severity` flag to `plugincheck2`.
This flag is used to set the severity of every rule in every analyzer and it has higher priority compared to all possible was set  severity by config and by the code.

### DIFF
--- a/pkg/cmd/plugincheck2/main.go
+++ b/pkg/cmd/plugincheck2/main.go
@@ -51,6 +51,11 @@ func main() {
 			"",
 			"Run a specific analyzer",
 		)
+		severity = flag.String(
+			"severity",
+			"",
+			"Set severity of the analyzer",
+		)
 	)
 
 	flag.Parse()
@@ -62,6 +67,7 @@ func main() {
 	logme.Debugln("archive file: ", flag.Arg(0))
 	logme.Debugln("checksum: ", *checksum)
 	logme.Debugln("analyzer: ", *analyzer)
+	logme.Debugln("severity: ", *severity)
 
 	cfg, err := readConfigFile(*configFlag)
 	if err != nil {
@@ -131,6 +137,7 @@ func main() {
 			ArchiveCalculatedSHA1: fmt.Sprintf("%x", sha1hash),
 		},
 		cfg,
+		analysis.Severity(*severity),
 	)
 	if err != nil {
 		logme.Errorln(fmt.Errorf("check failed: %w", err))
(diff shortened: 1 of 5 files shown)


pkg/cmd/plugincheck2/main.go
```diff
@@ -51,6 +51,11 @@ func main() {
 			"",
 			"Run a specific analyzer",
 		)
+		severity = flag.String(
+			"severity",
+			"",
+			"Set severity of the analyzer",
```
academo, line 57:
```suggestion
			"Set severity of the analyzer. Only works in combination with -analyzer",
```

pkg/runner/runner.go
```diff
@@ -146,6 +147,10 @@ func initAnalyzers(analyzers []*analysis.Analyzer, cfg *Config, pluginId string)
 				}
 			}
 
+			if severity != "" {
```
academo, line 150: This will overwrite the severity of every analyzer but that's not what we want. We only want to overwrite the one passed as a flag.

Imagine we pass severity here as "ok" but there are errors in the archive because the content is corrupted (could not unzip) or is not even a plugin. the validator will pass with `ok` for all of them.

We should make sure we are only overwriting the passed analyzer (close to line 121) and only for the one that was passed via flags

academo (changes requested):
this should be `analyzerSeverity` that affects the severity of the analyzer passed as a flag and not a global `-severity` that affect the global severity

pkg/runner/runner.go
oshirohugo: That's partially right.
This will overwrite the severity of every analyzer if we don't provide the `analizer` flag to select just one analyzer. If we do the severity of only the selected analyzer will be affected, and the ones from the required analyzer will remain as default.
E.g. If I have a package that doesn't have `plugin.json` and I run:
```
plugincheck2 -analyzer=metadatavalid -severity=ok ~/Work/wip/xx-aa-datasource/xx-aa-datasource-1.0.0.zip
``` 
as `metadatavalid` requires `metadata` and `metadataschema` I will get:

```
error: missing plugin.json
detail: A plugin.json file is required to describe the plugin.
ok: plugin.json not found
detail: plugin.json not found in the archive. Please refer to the documentation for more information. https://grafana.com/docs/grafana/latest/developers/plugins/metadata/
```
which shows that only the `metadatavalid` is affected by the `severity` while the required analyzers keep their default severities.

Therefore the easiest way to implement what you want (set severity only for a specific analyzer provided by `analyzer` flag) is to keep the same logic and only consider the `severity` flag (that I will change to `analyzerSeverity`) when `analyzer` is set.

academo (approved):

PR #275, opened 2024-11-07
Generate analyzers table in README

xnyo:
Introduces a new property to the `Analyzers` called `ReadmeInfo`. This property contains information for the README file. This information is then used to generate the table in the README.md file.

This moves the README information next to the analyzer's code, which makes it harder to forget to document what the analyzers actually do.

Example usage:

```go
var Analyzer = &analysis.Analyzer{
	// ...
	ReadmeInfo: analysis.ReadmeInfo{
		Name:        "Archive Name",
		Description: "The name of the archive should be correctly formatted.",
	},
}
```

To re-generate the README, run:

```bash
mage gen:readme
```

This PR also introduces a test case that compares the content of the README in the repo with the generated one. If they don't match, the test fails. This will make CI fail if the README is not re-generated.


pkg/genreadme/genreadme.go
```diff
@@ -0,0 +1,91 @@
...
+	"slices"
+	"strings"
+
+	"github.com/grafana/plugin-validator/pkg/analysis"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes"
+	"github.com/grafana/plugin-validator/pkg/logme"
+)
+
+const (
+	magicStart = `<!-- analyzers-table-start -->`
```
academo, line 17: image

academo (approved):
really nice work on this! it was so hard to keep the table up to date by hand

PR #278, opened 2024-11-13
feat: check screenshot image type

### LINKED ISSUE #277: Validate screenshot mime types
The `screenshots` validator ([here](https://github.com/grafana/plugin-validator/blob/main/pkg/analysis/passes/screenshots/screenshots.go)) should be updated to include a check of the mime type of the screenshots included in the zip file. Currently, it only checks if the files exist.

GCOM validates the screenshots mime types when publishing a plugin. An invalid or unsupported mime type will cause the following error:

```json
{
  "code": "InvalidArgument",
  "message": "Image is of unsupported type.",
  "requestId": "d3438a80-ea2c-49c1-94f9-ebea30afcad2"
}
```

Valid mime types are defined here:

https://github.com/grafana/grafana-com/blob/b45bb5f67d851018a0a3208720cc3fd5240edbf3/packages/grafana-com-lib-api/src/images/validation.ts#L105

The validator should check that the screenshots are supported by GCOM, or it will reject the plugin upon attempting to publish it.

s4kh:
Fixes https://github.com/grafana/plugin-validator/issues/277


pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -39,24 +44,59 @@ func checkScreenshotsExist(pass *analysis.Pass) (interface{}, error) {
...
+	}
 
-		if reportCount > 0 {
-			return nil, nil
+	reportCount := 0
+	for _, screenshot := range data.Info.Screenshots {
+		if strings.TrimSpace(screenshot.Path) == "" {
+			reportCount++
+			pass.ReportResult(pass.AnalyzerName, screenshots, fmt.Sprintf("plugin.json: invalid empty screenshot path: %q", screenshot.Name), "The screenshot path must not be empty.")
+		} else if !validImageType(screenshot.Path) {
```
s4kh, line 54: Is it better to do this after `invalid empty screenshot` in a separate loop? Added here so we can use the same loop.

pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -39,24 +49,59 @@ func checkScreenshotsExist(pass *analysis.Pass) (interface{}, error) {
...
-		if reportCount > 0 {
-			return nil, nil
+	reportCount := 0
+	for _, screenshot := range data.Info.Screenshots {
+		if strings.TrimSpace(screenshot.Path) == "" {
+			reportCount++
+			pass.ReportResult(pass.AnalyzerName, screenshots, fmt.Sprintf("plugin.json: invalid empty screenshot path: %q", screenshot.Name), "The screenshot path must not be empty.")
+		} else if !validImageType(filepath.Join(archiveDir, screenshot.Path)) {
+			reportCount++
+			pass.ReportResult(pass.AnalyzerName, screenshotsType, fmt.Sprintf("invalid screenshot image type: %q. Accepted image types: %q", screenshot.Path, acceptedImageTypes), "The screenshot image type invalid.")
```
academo, line 61: This error can be confusing for users if for example they have a typo in their plugin json or the wrong path. We'll be telling them the image type is wrong but in reality the image path is wrong.

Let's have an error if the image path can't be found and a separate one for the image type.

pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -39,24 +49,59 @@ func checkScreenshotsExist(pass *analysis.Pass) (interface{}, error) {
...
+		pass.ReportResult(pass.AnalyzerName, screenshotsType, "screenshots are valid image type", "")
 	}
 
 	return data.Info.Screenshots, nil
 }
+
+func validImageType(imgPath string) bool {
+	file, err := os.Open(imgPath)
+	if err != nil {
+		fmt.Printf("cannot open file: %v\n", err)
```
academo, line 85: don't output to stdout it'll break the validator output. use `logme.DebugFln`

pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -39,24 +49,59 @@ func checkScreenshotsExist(pass *analysis.Pass) (interface{}, error) {
...
+		return false
+	}
+	defer file.Close()
+
+	// 512 is enough for getting the content type
+	// https://pkg.go.dev/net/http#DetectContentType
+	buffer := make([]byte, 512)
+	if _, err := file.Read(buffer); err != nil {
+		fmt.Printf("cannot read file: %v\n", err)
```
academo, line 94: don't output to stdout use our logging lib

academo (commented):
nit: there's no need to put 120kb images for testing data, we can use a small 1kb image (a grafana icon maybe?)

Since we are at that, can we have a table test with a test case for each valid image type?

s4kh:
> nit: there's no need to put 120kb images for testing data, we can use a small 1kb image (a grafana icon maybe?)
> 
> Since we are at that, can we have a table test with a test case for each valid image type?

With the table test, I was able to catch an SVG validation error, thanks @academo. :100:

pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -39,24 +53,73 @@ func checkScreenshotsExist(pass *analysis.Pass) (interface{}, error) {
...
-			return nil, nil
+	reportCount := 0
+	for _, screenshot := range data.Info.Screenshots {
+		if strings.TrimSpace(screenshot.Path) == "" {
+			reportCount++
+			pass.ReportResult(pass.AnalyzerName, screenshots, fmt.Sprintf("plugin.json: invalid empty screenshot path: %q", screenshot.Name), "The screenshot path must not be empty.")
+		} else if err := validateImage(filepath.Join(archiveDir, screenshot.Path)); err != nil {
+			reportCount++
+			logme.Debugln(err)
+			pass.ReportResult(pass.AnalyzerName, screenshotsType, err.Error(), "The screenshot image.")
```
xnyo, line 66: Nit: I would change the message slightly

```suggestion
			pass.ReportResult(pass.AnalyzerName, screenshotsType, err.Error(), "The screenshot image is of an unsupported format.")
```

xnyo (approved):
LGTM, awesome work! Only one small nit:

academo (approved):

Issue #279, opened 2024-11-13
The validation of `grafanaDependency` uses a regex pattern that is too strict


darrenjaneczek:
See:
- https://github.com/grafana/plugin-validator/blob/762a056c2e9bb85c20b9c14fb4aafd941d45b2e3/config/plugin.schema.json#L173

This pattern:
`^(<=|>=|<|>|=|~|\\^)?([0-9]+)(\\.[0-9x\\*]+)(\\.[0-9x\\*]+)?(\\s(<=|>=|<|=>)?([0-9]+)(\\.[0-9x]+)(\\.[0-9x]+))?(\\-[0-9]+)?$`
does not permit more complicated, but valid, semantic version ranges such as:
- `>= 10.4.10 < 11 || >= 11.1.0`

Options:
- [ ] Remove the pattern, and allow any string for this field
  - #280 
- [ ] Attempt to build a new mechanism to evaluate the validity of the semver range 
  - Perhaps a different kind of validation layer can be added to this that makes use of some javascript code and the semver library itself.

# Background:

While building an urgent patch for the GEL Plugin (https://github.com/grafana/gex-plugins)  for a customer POC, our build process spontaneously failed due to a change of `plugin-validator` where a strict pattern for `grafanaDependency` is required to pass. The gex-plugins currently use `>= 10.4.10 < 11 || >= 11.1.0` to specify the officially supported range. This is a valid semver, but the pattern used in the schema for `grafanaDependency` does not allow for spaces, multiple operators, or multiple ranges separated by `||`.

To work around this issue, I had to disable this check altogether. In order to proceed, I will need to know which path the `plugin-validator` is choosing to take so I can respond accordingly.

- A. Keep this regex
- B. Remove this regex
- C. Remove this regex for now and later replace it with true semver test

If the choice is A, I will have to alter the officially supported version range in gex-plugins, or find a way to disable this strict check without also disabling overall plugin-validation.

See:
- https://raintank-corp.slack.com/archives/C01C4K8DETW/p1731326942187249?thread_ts=1731093846.187719&cid=C01C4K8DETW

darrenjaneczek:
@mckn, I have taken @academo's advice and disabled the plugin schema analyzer so that our build would succeed. It seems our project has bigger problems than just the dependency expression -- our plugin id is also "in violation."

The solution I went with is here:
- https://github.com/grafana/plugin-validator/pull/280#issuecomment-2525238033

mckn:
[...]

PR #280, opened 2024-11-13
fix: remove strict pattern on grafanaDependency

darrenjaneczek:
Resolves part of:
- #279

### DIFF
--- a/config/plugin.schema.json
+++ b/config/plugin.schema.json
@@ -170,7 +170,6 @@
         "grafanaDependency": {
           "type": "string",
           "description": "Required Grafana version for this plugin. Validated using https://github.com/npm/node-semver.",
-          "pattern": "^(<=|>=|<|>|=|~|\\^)?([0-9]+)(\\.[0-9x\\*]+)(\\.[0-9x\\*]+)?(\\s(<=|>=|<|=>)?([0-9]+)(\\.[0-9x]+)(\\.[0-9x]+))?(\\-[0-9]+)?$"
         },
         "plugins": {
           "type": "array",


academo (changes requested):
You can disable specific validations for your plugin via a configuration file as described here https://github.com/grafana/plugin-validator?tab=readme-ov-file#using-a-configuration-file

The current condition, even though strict, still serves a meaningful purpose for most of the cases, removing it will open the possibility to plugins submitting wrong versions (that are not as complex as your specific case). Keep in mind we run the validator for community submissions and in the review process we can manually decide to overlook an specific problem such as this if the author explains the problem or we note it. 

Additionally, this file is a fallback for the validator, the validator always fetches the latest schema from the grafana/grafana repository https://github.com/grafana/grafana/blob/ad3fc957558022dc4199a6c060f06bf038c7d7bd/docs/sources/developers/plugins/plugin.schema.json#L

Alternatively maybe you can propose an updated regex that covers the condition instead of removing it?

darrenjaneczek:
> You can disable specific validations for your plugin via a configuration file as described here https://github.com/grafana/plugin-validator?tab=readme-ov-file#using-a-configuration-file

It is not clear how I can disable this specific check for the `grafanaDependency` entry.
Until I can, my only other options are to disable this check entirely, or change my version range to something simpler like `>=10.4.11` to comply.

> Additionally, this file is a fallback for the validator, the validator always fetches the latest schema from the grafana/grafana repository https://github.com/grafana/grafana/blob/ad3fc957558022dc4199a6c060f06bf038c7d7bd/docs/sources/developers/plugins/plugin.schema.json#L

That is definitely a problem here.
 
> Alternatively maybe you can propose an updated regex that covers the condition instead of removing it?

I don't believe a regex is a the best tool for parsing a valid semver version range. It would be difficult (if possible?) to derive an exhaustive pattern that covers all cases, and it would be repetitious and hard to read as a human. I gave it a time-boxed try, and failed to find something satisfactory.

In #279, I suggest an alternative check which makes use of the semver library itself. I don't think that's an option though if we're just checking through the schema.

darrenjaneczek:
@academo, I have closed this PR since based on what you said it doesn't disable the pattern (due to it being expressed in `grafana/grafana`). This is still a unresolved problem for our project. See #279 for more detail.

>You can disable specific validations for your plugin via a configuration file as described here https://github.com/grafana/plugin-validator?tab=readme-ov-file#using-a-configuration-file

Looking at the documentation, it is not clear how I can disable this specific check for the `grafanaDependency` entry. Any help would be appreciated.

darrenjaneczek:
Disabling the analyzers seems to be the best solution for our project:

```
global:
  enabled: true
  severity: warning
  jsonOutput: false
  reportAll: false
analyzers:
  license:
    enabled: false
  # The metadatavalid metadataschema checkers are too strict for our plugin
  # Our grafanaDependency version range is complicated,
  # and our plugin ids have internal dashes.
  metadatavalid:
    enabled: false    
  metadataschema:
    enabled: false      
```

PR #292, opened 2024-12-16
Fix license check timeouts with some backend executables

### LINKED ISSUE #291: Investigate: Issue with iodb license file https://github.com/apache/iotdb-extras/blob/master/connectors/grafana-plugin/LICENSE
The license validator is failing to validate this specific one. 

investigate and fix what could be happening with this specific plugin

xnyo:
The license check times out with some plugins (Apache IoTDB) and the validator fails. This can happen if `go-license-detector` attempts to run the license checks on backend executables, which are big and cause a timeout.

The `go-license-detector` library does a preliminary filter on all files in the zip file using a regex:

immagine

The Apache IoTDB backend executable triggers the check, most likely because the word "apache" is somewhere in the binary. This causes the license check to time out when the validator is run against the Apache IoTDB plugin zip.

This PR changes the `filer.Filer` passed to `go-license-detector` to a custom one which adds an additional filter based on the file's MIME type. All files whose MIME type doesn't start with `text/` will be excluded from the license checks. In this way, backend executables are never passed to `go-license-detector` and the check doesn't time out anymore.

Fixes #291.


academo (approved):
could you please add a test for the binary case?

xnyo:
> could you please add a test for the binary case?

Added in ab2e5e2725357c0493acec448081d06ae6814230

academo (approved):

PR #302, opened 2025-01-22
Add error for `topnav` toggle

### LINKED ISSUE #92735: Comms around `topnav` being removed for g12
- this will primarily be sharing with plugins **early** in the G12 cycle (e.g. after G11 is released) what needs to be done:
  - for plugins with a minimum grafana version >=9.5.0, they can freely remove any check on `topnav` and just use the code that's present when it is `true`
  - for plugins with an older minimum version, they should remove any check on `topnav` and just use the code that's present when it is `true` **and** update the minimum grafana version to >=9.5.0

ashharrison90:
- the `topnav` toggle has been deprecated since grafana 10.2 or something
- adds a warning to remove any reliance on the `topnav` toggle

Fixes https://github.com/grafana/grafana/issues/92735


pkg/analysis/passes/coderules/semgrep-rules.yaml
```diff
@@ -141,3 +141,20 @@ rules:
...
+        - "src/**/*.ts"
+        - "src/**/*.tsx"
+      exclude:
+        - "*.spec.ts"
+        - "*.spec.tsx"
+        - "*.test.ts"
+        - "*.test.tsx"
+        - "*.js"
+    message: "The `topnav` toggle is deprecated and will be removed in a future version of Grafana. Plugins should default to using the code where the toggle is enabled."
+    languages: [javascript, typescript]
```
xnyo, line 159: Nit: `*.js` files are excluded (`paths.exclude` property), so this should be:

```suggestion
    languages:
      - typescript
```

xnyo (commented):
Great work! :raised_hands: I have only one small comment

pkg/analysis/passes/coderules/semgrep-rules.yaml
```diff
@@ -141,3 +141,20 @@ rules:
...
+        - "src/**/*.tsx"
+      exclude:
+        - "*.spec.ts"
+        - "*.spec.tsx"
+        - "*.test.ts"
+        - "*.test.tsx"
+        - "*.js"
+    message: "The `topnav` toggle is deprecated and will be removed in a future version of Grafana. Plugins should default to using the code where the toggle is enabled."
+    languages: [typescript]
+    severity: WARNING
```
academo, line 160: make it an error. that will notify authors they have to take action.

academo (approved):

PR #303, opened 2025-01-23
New validator: Validate artifact provenance attestation if coming from a github pipeline

academo:
Adds a validator to validate github assets provenance attestation.

Currently supports only [github provenance attestations](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries) which can be easily generated by the [plugins build action](https://github.com/grafana/plugin-actions/tree/main/build-plugin/#attestation-of-plugin-package)

waiting on https://github.com/grafana/plugin-tools/pull/1475 to merge so I can update the links.


pkg/cmd/plugincheck2/main.go
```diff
@@ -136,8 +152,10 @@ func main() {
 	diags, err := runner.Check(
 		analyzers,
 		analysis.CheckParams{
+			ArchiveFile:           tmpZip.Name(),
```
academo, line 155: to validate the provenance we need unaltered plugin archive information and the source code reference.

pkg/analysis/passes/provenance/provenance.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+var (
+	noProvenanceAttestation = &analysis.Rule{
+		Name:     "no-provenance-attestation",
+		Severity: analysis.Warning,
+	}
+	invalidProvenanceAttestation = &analysis.Rule{
+		Name:     "invalid-provenance-attestation",
+		Severity: analysis.Warning,
```
xnyo, line 23: Should this be an error instead? (are we setting it just as a warning for now so we don't block submissions)?

pkg/analysis/passes/provenance/provenance.go
```diff
@@ -0,0 +1,169 @@
...
+			attestationPipeline,
+		)
+	}
+
+	return nil, nil
+}
+
+func getGithubProvenanceAttestationPipeline(assetPath string, repo string) (string, error) {
+
+	ghCliBin, err := getGithubCliPath()
```
xnyo, line 106: I would prefer using a library of some sort rather than depending on the `gh` CLI command. Have you checked if something suitable exists? Was there another reason why `gh` was used directly?

Also, do we already have `gh` available in CI, when we run the community pipeline?

xnyo (commented):
Great work! :clap::clap:  Here are some initial comments:

pkg/analysis/passes/provenance/provenance.go
academo: I checked and github doens't have any API or library available to do this validation it seems the GH CLi is the only way to do it or the "offline" method that requires much more setup and download to work.

I'll later open a PR in the our community pipeline to install GH Cli

pkg/analysis/passes/provenance/provenance.go
academo: the problem is that we don't know if a provenance was generated or not, we only know that is not valid.

if we make this an error we'll start failing all github-stored packages

pkg/analysis/passes/provenance/provenance.go
academo: I was mistaken I can indeed use the API of github for this. I updated the code

xnyo (commented):
Great work! :clap: :clap: :clap: 

Just a few comments but everything else looks great!

pkg/analysis/passes/provenance/provenance.go
xnyo: Closing as resolved :+1:

pkg/analysis/passes/provenance/provenance.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, nil
+	}
+
+	if noProvenanceAttestation.ReportAll {
+		noProvenanceAttestation.Severity = analysis.OK
+		pass.ReportResult(
+			pass.AnalyzerName,
+			noProvenanceAttestation,
+			"Provenance attestation found",
+			"", // TODO add details
```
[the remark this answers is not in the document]
academo: no. I actually forgot to add the details :D thanks

xnyo (approved):
LGTM! :rocket:

PR #317, opened 2025-03-31
LLM Validator: Improve questions and prompt generation

academo:
### main change
It now asks the model one question at the time instead of all questions at once.

This change allows the model to focus on replying a single question which greatly improves its answers. It also means the validator now consumes more api credits from google when running. 

### Other changes

* Updates the LLM model to use `gemini-2.0-flash-001` (stable gemini flash 2)
* Changes the question structure to have an expected answer instead of always expecting a positive
* Adds structured output to LLM
* Adds more questions for code
* Moves retrial logic to per-question

### DIFF
--- a/pkg/llmvalidate/llmvalidate.go
+++ b/pkg/llmvalidate/llmvalidate.go
@@ -13,9 +13,8 @@ import (
 
 	"github.com/danwakefield/fnmatch"
 	"github.com/google/generative-ai-go/genai"
-	"google.golang.org/api/option"
-
 	"github.com/grafana/plugin-validator/pkg/logme"
+	"google.golang.org/api/option"
 )
 
 // these are not regular expressions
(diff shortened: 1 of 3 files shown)


pkg/llmvalidate/llmvalidate.go
```diff
@@ -86,15 +89,30 @@ type Client struct {
...
+
 func New(ctx context.Context, apiKey string, modelName string) (*Client, error) {
 
 	if apiKey == "" {
 		return nil, fmt.Errorf("API key is required")
 	}
 
 	if modelName == "" {
-		modelName = "gemini-1.5-flash-latest"
+		modelName = "gemini-2.0-flash-lite"
```
academo, line 113: This intentionally uses the `lite` model. it is a fallback in case no model is passed

xnyo: Sounds good, but it feels a bit weird to me that the model name (`gemini-2.0-flash-001`) is defined in this file as well as in `pkg/analysis/passes/llmreview/llmreview.go` ([here](https://github.com/grafana/plugin-validator/blob/2de8c89e73419d30804d3c878f45f44d1ccb29a9/pkg/analysis/passes/llmreview/llmreview.go#L86)).

I would keep all the model names in this file as constants, and use them instead of using strings:

```go
package llmvalidate

// ...

const (
	ModelGemini2_0FlashLite = "gemini-2.0-flash-lite"
	ModelGemini2_0Flash001  = "gemini-2.0-flash-001"
)
```

```suggestion
		modelName = ModelGemini2_0FlashLite
```

and in `pkg/analysis/passes/llmreview/llmreview.go`

```go
llmClient, err := llmvalidate.New(context.Background(), geminiKey, llmvalidate.ModelGemini2_0Flash001)
```

Or have you checked if the model names are already defined in `github.com/google/generative-ai-go/genai` ?

xnyo (approved):
I haven't tested this but the code LGTM apart from a few minor nitpicks. Awesome work! :raised_hands:

pkg/llmvalidate/llmvalidate.go
academo: They are not included in the library. I removed the fallback for the model and kept only this string.

PR #320, opened 2025-04-15
Make license validator work with a single license file to prevent library oddities

### LINKED ISSUE #293: License validator: simplify logic to not rely on file system scanning
From https://github.com/grafana/plugin-validator/issues/291

The current code and library are doing file system scans to retrieve licenses but we know that the license should be called `LICENSE` and bit on the root of the plugin archive.

We can simplify the existing logic to those constraints and remove all the library calls that use the file system.

academo:
Modifies the validator so instead of passing the full archive contents to the license library it passes a new temporal directory only with a license file.

This prevents the license library to scan all the files in the archive and potentially detect false positives or get locked on scanning big files

Closes https://github.com/grafana/plugin-validator/issues/293

### DIFF
--- a/pkg/analysis/passes/license/license.go
+++ b/pkg/analysis/passes/license/license.go
@@ -1,27 +1,21 @@
 package license
 
 import (
-	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
-	"time"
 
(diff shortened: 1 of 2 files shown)


oshirohugo (approved):

xnyo (approved):
LGTM!

PR #323, opened 2025-04-30
Security: Fix zizmur issues

academo:


.github/workflows/do-release.yml
```diff
@@ -15,26 +15,40 @@ on:
 jobs:
   bump-version:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
academo, line 19: cloning the repository with the app token "should" be enough to modify the files. we will have to test it and if not put back `write` in contents here

andresmgot (approved):

xnyo (approved):
LGTM!

PR #331, opened 2025-05-07
Add recommendation to specify sponsorship link

Ukochka:
We would like to encourage plugin developers to consider adding a sponsorship link in their plugin.json file, which will be shown on the plugin details page to allow users to support their work if they wish.


pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	for _, link := range data.Info.Links {
+		name := strings.ToLower(link.Name)
+		if strings.Contains(name, "sponsor") || strings.Contains(name, "sponsorship") {
+			return nil, nil
+		}
+	}
+	explanation := "Consider to add a sponsorship link in your plugin.json file (Info.Links section), which will be shown on the plugin details page to allow users to support your work if they wish."
+	pass.ReportResult(pass.AnalyzerName, sponsorshiplink, "plugin.json: You can include a sponsorship link if you want users to support your work", explanation)
```
academo, line 51:
```suggestion
	pass.ReportResult(pass.AnalyzerName, sponsorshiplink, "You can include a sponsorship link if you want users to support your work", explanation)
```

no need to add `plugin.json:` in these messages

pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, nil
+	}
+
+	for _, link := range data.Info.Links {
+		name := strings.ToLower(link.Name)
+		if strings.Contains(name, "sponsor") || strings.Contains(name, "sponsorship") {
+			return nil, nil
+		}
+	}
+	explanation := "Consider to add a sponsorship link in your plugin.json file (Info.Links section), which will be shown on the plugin details page to allow users to support your work if they wish."
```
academo, line 50: I am not a fan of a default fail behaviour. it makes hard to understand when it should fail and when it should pass.

I recommend you to re-do the logic so the report only happens when a fail condition is met instead of when a success condition is not met.

academo, line 50: duplicated text. declare it only once so we only have to edit it once in the future.

pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
```diff
@@ -0,0 +1,54 @@
...
+	Run:      checkSponsorshiplink,
+	Requires: []*analysis.Analyzer{metadata.Analyzer, archive.Analyzer},
+	Rules:    []*analysis.Rule{sponsorshiplink},
+	ReadmeInfo: analysis.ReadmeInfo{
+		Name:        "Sponsorship Link",
+		Description: "Checks if a sponsorship link is specified in `plugin.json` that will be shown in the Grafana plugin catalog for users to support the plugin developer.",
+	},
+}
+
+func checkSponsorshiplink(pass *analysis.Pass) (interface{}, error) {
```
academo, line 29:
```suggestion
func run(pass *analysis.Pass) (interface{}, error) {
```

pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
```diff
@@ -0,0 +1,54 @@
...
+	metadataBody, ok := pass.ResultOf[metadata.Analyzer].([]byte)
```
academo, line 30: do plugins also put `links` in nested plugins plugin.json? if so you should be using `nestedmetadata` and checking every plugin json. e.g. https://github.com/grafana/plugin-validator/blob/15a7d6eb85255221cd48cf9aeff7a9595ac6cdba/pkg/analysis/passes/logos/logos.go#L8

Ukochka: I think it is not necessary to handle this case here. We add this rule temporary just to show developers that they can configure such a link

.github/workflows/do-release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
academo, line 37: are these changes part of your PR?

Ukochka: No, it was not changed. only line 40 was changed by you

pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+	"strings"
+
+	"github.com/grafana/plugin-validator/pkg/analysis"
+	"github.com/grafana/plugin-validator/pkg/analysis/passes/metadata"
+)
+
+var (
+	sponsorshiplink = &analysis.Rule{Name: "sponsorshiplink", Severity: analysis.Recommendation}
+	explanation     = "Consider to add a sponsorship link in your plugin.json file (Info.Links section), which will be shown on the plugin details page to allow users to support your work if they wish."
```
xnyo, line 13: I'd also add they should use "sponsor" or "sponsorship" as the link name otherwise it won't be picked up

pkg/analysis/passes/sponsorshiplink/sponsorship_test.go
xnyo, line 1: Nit: This file should be renamed to `sponsorshiplink_test.go`

xnyo (commented):
Great work! Some comments, you should also rebase your PR because for some reason there are some changes that are not related to your work (e.g.: dependency bumps, license analyzer changes, semgrep changes) which come from other recent PRs

pkg/analysis/passes/sponsorshiplink/sponsorshiplink.go
academo: if this is temporary, please open an issue with a due date to remove it

xnyo (approved):
LGTM, great work! :rocket:

PR #336, opened 2025-05-23
Add webrisk validation for links

Ukochka:
This rule validates that link is safe using web risk API https://cloud.google.com/web-risk/docs/reference/rest/v1/uris/search

All other checks for working link has been validated in brokenlinks rule

Next step:
Create an API key to access this endpoint


pkg/analysis/passes/safelinks/safelinks.go
```diff
@@ -0,0 +1,120 @@
+package safelinks
```
academo, line 1: please add tests for this.

you will have to mock the webrisk api because we don't want to hit the API when running tests.

pkg/analysis/passes/safelinks/safelinks.go
```diff
@@ -0,0 +1,164 @@
...
+		result := LinkResult{Link: link}
+
+		if link.URL == "" {
+			results[i] = result
+			continue
+		}
+
+		params := url.Values{}
+		params.Set("uri", link.URL)
+		params.Set("key", webriskApiKey)
```
academo, line 104: I don't like the idea of us sending the key over the http parameters of a get request. this can easily be logged in many places.

Check if we can use the header `X-Goog-Api-Key` instead as documented here https://cloud.google.com/docs/authentication/api-keys-use#using-with-rest

pkg/analysis/passes/safelinks/safelinks.go
```diff
@@ -0,0 +1,164 @@
...
+		params.Add("threatTypes", string(ThreatTypeSocialEngineeringExtendedCoverage))
+
+		apiURL := fmt.Sprintf("%s?%s", webRiskAPIBaseURL, params.Encode())
+
+		reqCtx, cancel := context.WithTimeout(ctx, requestTimeout)
+		req, err := http.NewRequestWithContext(reqCtx, "GET", apiURL, nil)
+		if err != nil {
+			cancel()
+			result.Error = fmt.Errorf("failed to create request: %w", err)
+			results[i] = result
```
academo, line 117: not sure if you need to assign results[i] = result every time or you can do this just at the start of the loop (line 95)

pkg/analysis/passes/safelinks/safelinks_test.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	_, err := Analyzer.Run(pass)
+	require.NoError(t, err)
+	require.Empty(t, interceptor.Diagnostics)
+}
+
+func TestSafeLink(t *testing.T) {
+	webriskApiKey := os.Getenv("WEBRISK_API_KEY")
+	require.NotEmpty(t, webriskApiKey, "API key should not be empty")
```
academo, line 47: this will make the test fail in build time for environments that don't define it. (e.g. some local environments)

What we've done with other tests is to skip the test if the env is not defined.

xnyo (commented):
Great work! :tada: Some comments:

pkg/analysis/passes/safelinks/safelinks.go
academo: @xnyo thanks for that! I had the impression you could only do that with an array of pointers but was not sure.

pkg/analysis/passes/safelinks/safelinks.go
```diff
@@ -0,0 +1,162 @@
...
+	return results
+}
+
+func (t threatTypes) String() string {
+	threatTypeStrings := make([]string, len(t))
+	for i, threatType := range t {
+		threatTypeStrings[i] = string(threatType)
+	}
+	return strings.Join(threatTypeStrings, ", ")
+}
```
xnyo, line 162: Nit: This is a method of `threatTypes` and should be moved right after to the definition of the `threatTypes` type, so around line 22. Not a requirement but good practice :slightly_smiling_face:

xnyo (commented):
Awesome work! :tada:  A couple of minor comments to address but the rest looks good!

pkg/analysis/passes/safelinks/safelinks_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+func TestNoAPIKey(t *testing.T) {
+	// Temporarily unset the API key
+	originalKey := os.Getenv("WEBRISK_API_KEY")
+	os.Unsetenv("WEBRISK_API_KEY")
+	defer func() {
+		if originalKey != "" {
+			os.Setenv("WEBRISK_API_KEY", originalKey)
+		}
+	}()
```
xnyo, line 27: Nit: It's better to use `t.Cleanup` in tests rather than `defer`. Also, `os.Setenv` and `os.Unsetenv` return an `error` and it should be checked:

```suggestion
	// Temporarily unset the API key
	originalKey := os.Getenv("WEBRISK_API_KEY")
	err := os.Unsetenv("WEBRISK_API_KEY")
	require.NoError(t, err)
	t.Cleanup(func() {
		if originalKey != "" {
			err := os.Setenv("WEBRISK_API_KEY", originalKey)
			require.NoError(t, err)
		}
	})
```

pkg/analysis/passes/safelinks/safelinks_test.go
xnyo, line 1: I had a better look at the tests and they are always skipped unless `WEBRISK_API_KEY` is set, however the requests are mocked.

The tests should not be skipped depending on that env var, or tests won't run in CI. The check for the env var inside the test is not needed because the requests to the webrisk api are mocked, and we never hit the real webrisk API.

The following changes to the test file make the test pass regardless of the `WEBRISK_API_KEY` variable:

```diff
diff --git a/pkg/analysis/passes/safelinks/safelinks_test.go b/pkg/analysis/passes/safelinks/safelinks_test.go
index 4c6ed57..6f6d8d2 100644
--- a/pkg/analysis/passes/safelinks/safelinks_test.go
+++ b/pkg/analysis/passes/safelinks/safelinks_test.go
@@ -14,17 +14,32 @@ import (
        "github.com/grafana/plugin-validator/pkg/testpassinterceptor"
 )
 
-const pluginId = "test-plugin-panel"
+const (
+       pluginId = "test-plugin-panel"
+       envName  = "WEBRISK_API_KEY"
+)
+
+func setAPIKey(t *testing.T) {
+       old := os.Getenv(envName)
+       err := os.Setenv(envName, "test-api-key")
+       require.NoError(t, err)
+       t.Cleanup(func() {
+               err := os.Setenv(envName, old)
+               require.NoError(t, err)
+       })
+}
 
 func TestNoAPIKey(t *testing.T) {
        // Temporarily unset the API key
-       originalKey := os.Getenv("WEBRISK_API_KEY")
-       os.Unsetenv("WEBRISK_API_KEY")
-       defer func() {
+       originalKey := os.Getenv(envName)
+       err := os.Unsetenv(envName)
+       require.NoError(t, err)
+       t.Cleanup(func() {
                if originalKey != "" {
-                       os.Setenv("WEBRISK_API_KEY", originalKey)
+                       err := os.Setenv(envName, originalKey)
+                       require.NoError(t, err)
                }
-       }()
+       })
 
        var interceptor testpassinterceptor.TestPassInterceptor
        pass := &analysis.Pass{
@@ -37,18 +52,13 @@ func TestNoAPIKey(t *testing.T) {
                Report: interceptor.ReportInterceptor(),
        }
 
-       _, err := Analyzer.Run(pass)
+       _, err = Analyzer.Run(pass)
        require.NoError(t, err)
        require.Empty(t, interceptor.Diagnostics)
 }
 
 func TestSafeLink(t *testing.T) {
-       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
-       if webriskApiKey == "" {
-               t.Skip("API key not set, skipping test")
-               return
-       }
-
+       setAPIKey(t)
        httpmock.ActivateNonDefault(httpClient)
        defer httpmock.DeactivateAndReset()
 
@@ -79,12 +89,7 @@ func TestSafeLink(t *testing.T) {
 }
 
 func TestMalwareLink(t *testing.T) {
-       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
-       if webriskApiKey == "" {
-               t.Skip("API key not set, skipping test")
-               return
-       }
-
+       setAPIKey(t)
        httpmock.ActivateNonDefault(httpClient)
        defer httpmock.DeactivateAndReset()
 
@@ -121,11 +126,7 @@ func TestMalwareLink(t *testing.T) {
 }
 
 func TestSocialEngineeringLink(t *testing.T) {
-       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
-       if webriskApiKey == "" {
-               t.Skip("API key not set, skipping test")
-               return
-       }
+       setAPIKey(t)
        httpmock.ActivateNonDefault(httpClient)
        defer httpmock.DeactivateAndReset()
 
@@ -164,12 +165,7 @@ func TestSocialEngineeringLink(t *testing.T) {
 }
 
 func TestMultipleThreatTypes(t *testing.T) {
-       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
-       if webriskApiKey == "" {
-               t.Skip("API key not set, skipping test")
-               return
-       }
-
+       setAPIKey(t)
        httpmock.ActivateNonDefault(httpClient)
        defer httpmock.DeactivateAndReset()
 
@@ -208,12 +204,7 @@ func TestMultipleThreatTypes(t *testing.T) {
 }
 
 func TestMultipleLinks(t *testing.T) {
-       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
-       if webriskApiKey == "" {
-               t.Skip("API key not set, skipping test")
-               return
-       }
-
+       setAPIKey(t)
        httpmock.ActivateNonDefault(httpClient)
        defer httpmock.DeactivateAndReset()
```

- Removed the `WEBRISK_API_KEY` env var checks
- Added a `setAPIKey` function to set the env var value to a dummy one, so the analyzer runs (with mocked HTTP requests)
- Call this function at the top of each test that used mocked HTTP requests

You will also need those changes in `pkg/analysis/passes/safelinks/safelinks.go`:

```diff
diff --git a/pkg/analysis/passes/safelinks/safelinks.go b/pkg/analysis/passes/safelinks/safelinks.go
index 9afd992..63bdc88 100644
--- a/pkg/analysis/passes/safelinks/safelinks.go
+++ b/pkg/analysis/passes/safelinks/safelinks.go
@@ -52,8 +52,6 @@ var allThreatTypes = []threatType{
        threatTypeSocialEngineeringExtendedCoverage,
 }
 
-var webriskApiKey = os.Getenv("WEBRISK_API_KEY")
-
 var (
        webriskFlagged    = &analysis.Rule{Name: "webrisk-flagged", Severity: analysis.Error}
        webRiskAPIBaseURL = "https://webrisk.googleapis.com/v1/uris:search"
@@ -73,6 +71,7 @@ var Analyzer = &analysis.Analyzer{
 }
 
 func run(pass *analysis.Pass) (interface{}, error) {
+       webriskApiKey := os.Getenv("WEBRISK_API_KEY")
        metadataBody, ok := pass.ResultOf[metadata.Analyzer].([]byte)
        if !ok {
                return nil, nil
@@ -93,7 +92,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
        ctx := context.Background()
 
-       results := checkURLs(ctx, data.Info.Links)
+       results := checkURLs(ctx, webriskApiKey, data.Info.Links)
 
        for _, result := range results {
                if result.err != nil {
@@ -109,7 +108,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
        return nil, nil
 }
 
-func checkURLs(ctx context.Context, links []metadata.Link) []linkResult {
+func checkURLs(ctx context.Context, apiKey string, links []metadata.Link) []linkResult {
        results := []linkResult{}
 
        for _, link := range links {
@@ -137,7 +136,7 @@ func checkURLs(ctx context.Context, links []metadata.Link) []linkResult {
                        continue
                }
 
-               req.Header.Set("X-goog-api-key", webriskApiKey)
+               req.Header.Set("X-goog-api-key", apiKey)
 
                resp, err := httpClient.Do(req)
```

This is because we need to evaluate `webriskApiKey` (so get the value of the env var) each time we call `run` (for each test), instead of only once at package-level (otherwise it's always an empty string).

xnyo (commented):
Thank you for addressing the previous comments! I had another look at the tests and they should be changed slightly. I'm also about to test with a real api key so I may leave more comments in case I find other issues down the line

xnyo (approved):
Awesome work! :tada: I have tested it locally as well and it works. I only have a small comment regarding logging, but LGTM otherwise!

PR #346, opened 2025-06-27
fix(screenshots): handle malformed screenshots format gracefully

tolzhabayev:
- Metadata and nestedmetadata validators will not error when the metadata is invalid. instead they'll return nil which will cause others validators to not run. [1]
- Add proper error handling for malformed screenshots JSON format
- Previously, array of strings would cause validator to crash with unmarshaling error
- Now reports clear validation error with helpful message explaining correct format
- Add new 'screenshots-format' rule to catch and report format violations
- Include comprehensive test case for malformed format handling

[1] This guarantees that other validators that don't require metadata run and allow us to report correctly schema validation issues instead of failing the pipeline as currently it is happening. 

Fixes issue where plugins with screenshots as array of strings would crash validator instead of reporting a proper validation error. Example of a crash https://github.com/grafana/community-pipeline/actions/runs/15917761529/job/44898483257#step:5:127

### DIFF
--- a/pkg/analysis/passes/metadata/metadata.go
+++ b/pkg/analysis/passes/metadata/metadata.go
@@ -1,6 +1,7 @@
 package metadata
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 
@@ -47,6 +48,12 @@ func run(pass *analysis.Pass) (interface{}, error) {
 			}
 		}
 		return nil, err
+	}
+	var data Metadata
+	if err := json.Unmarshal(b, &data); err != nil {
+		// if we fail to unmarshall it means the schema is incorrect
+		// we will let the metadatavaid validator handle it
+		return nil, nil
 	}
 
 	return b, nil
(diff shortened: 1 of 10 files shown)


wbrowne (approved):
LGTM

pkg/analysis/passes/screenshots/screenshots.go
```diff
@@ -46,6 +47,13 @@ func checkScreenshots(pass *analysis.Pass) (interface{}, error) {
 
 	var data metadata.Metadata
 	if err := json.Unmarshal(metadataBody, &data); err != nil {
+		// Check if the error is related to screenshots field format
```
academo, line 50: did we encounter an error here before? the reason this code is not present anywhere where we use the matadata is because this validator depends on `metadata.Analyzer`. if there's an error in the metadata json format it'll fail there and not here. 

in other words, we are guarantee to t least have a correct json here. not necessarily the correct schema, but correct json.

academo (approved):

xnyo (approved):
LGTM!

pkg/analysis/passes/metadata/metadata.go
```diff
@@ -48,6 +49,12 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		}
 		return nil, err
 	}
+	var data Metadata
```
academo, line 52: previously we didn't check if we could unmarshall the metadata before. this led to other validators blindly trying to unmarshall the metadata and getting errors

PR #368, opened 2025-09-01
Codediff: Add gemini cli invokation to review code diff if available

academo:
Adds GEMINI CLI integration to the code-diff if a version diff for the current plugin is available.

The code-diff review will use the same questions from the existing LLMReview but applies them only to the code diff between version releases.


pkg/analysis/passes/codediff/codediff.go
```diff
@@ -99,3 +162,97 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		return nil, err
+	}
+
+	// clean up files from repositoryPath
+	cleanFiles := []string{"replies.json", ".nvmrc", "GEMINI.MD"}
+	for _, file := range cleanFiles {
+		filePath := filepath.Join(repositoryPath, file)
+		if _, err := os.Stat(filePath); err == nil {
+			if err := os.Remove(filePath); err != nil {
+				logme.Debugln("Failed to remove file:", err)
```
xnyo, line 225: Nit: Maybe use warning-level rather than debug, since we are also swallowing the error?

pkg/analysis/passes/codediff/codediff.go
```diff
@@ -99,3 +162,97 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	if err != nil {
+		logme.Debugln("Failed to generate prompt:", err)
+		return nil, err
+	}
+
+	// clean up files from repositoryPath
+	cleanFiles := []string{"replies.json", ".nvmrc", "GEMINI.MD"}
+	for _, file := range cleanFiles {
+		filePath := filepath.Join(repositoryPath, file)
+		if _, err := os.Stat(filePath); err == nil {
```
xnyo, line 223: The err returned by `os.Stat` should be handled as well

pkg/analysis/passes/codediff/prompt.txt
```diff
@@ -0,0 +1,30 @@
+You will review the changes in this Grafana plugin. Use `git diff` see the diff and `git show` to see full files in an specific commit. You should generally use the git tool as necessary. 
+
+Make sure to analyze the file changes in the context of its original version to understand well what the changes do
+
+Analyze the changes introduced in between the versions, focus on go, ts and tsx files, ignore test files, dev files and hidden files. Analyze only on the changes introduced in between versions. if a change looks significant, make sure to analyze it in detail comparing full file versions if necessary.
```
xnyo, line 5: Nit:

```suggestion
Analyze the changes introduced in between the versions, focus on go, ts and tsx files. Ignore test files, dev files and hidden files. Analyze only on the changes introduced in between versions. if a change looks significant, make sure to analyze it in detail comparing full file versions if necessary.
```

xnyo (commented):
Awesome work! I don't have a Gemini API key to test this out, but the code changes make sense! I only have a few small comments:

pkg/analysis/passes/codediff/codediff.go
academo: yes. this is an expected error. but we don't want to print it so we swallow it

xnyo (approved):
LGTM! :raised_hands:

pkg/analysis/passes/codediff/codediff.go
academo: This error will trigger very often so I prefer we don't log it to reduce noise

academo:
> > The code-diff review will use the same questions from the existing LLMReview but applies them only to the code diff between version releases.
> 
> I wonder if we are just going to get duplicated responses here? Do you have an example execution?

I did many tests and I didn't get duplicated answers. Mainly because we are asking the LLM in `llmreview` to review all the code and the llm here `codediff` to review the changes in between versions.

In an idea world the LLM will only report the answer to these questions based on the git diff not on the whole code but LLMs will LLMs and it might happen from time to time.

hugohaggmark (commented):
I've tried it out locally and it works as I would expect. Would be nice to have a test plugin where we know in advance what answers to expect.

hugohaggmark (approved):
Works as expected 🎉

PR #369, opened 2025-09-08
flags: add option to output validator json output to file

academo:
This will allow the validator to write its output to a file instead of only to stdout


xnyo (approved):
LGTM!

PR #373, opened 2025-09-11
chore: clean up unnecessary report all (part 2)

academo:
* Makes sure all rules are defined in each analyzer metadata
* Cleans up some unnecessary report all
* Adds some better reportAll OK Checks


briangann (approved):
looks great!

PR #381, opened 2025-09-18
coderules: validate no react internals are accessed

### LINKED ISSUE #350: Report on incompatible react libraries
It would be useful if we could validate if the plugins bundled JS files contain strings related to usage of react internals so we can start getting plugins React 19 compatible.

To update Grafana core to React 19 we need plugins to be compatible with react 19. The major issue we're facing right now is that plugins bundle react components / libraries that are incompatible with React 19 due to usage of the internal react APIs. In React 18 it's named  `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` however it's been renamed to `__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE`.

academo:
closes https://github.com/grafana/plugin-validator/issues/350

### DIFF
--- a/pkg/analysis/passes/coderules/coderules_test.go
+++ b/pkg/analysis/passes/coderules/coderules_test.go
@@ -232,3 +232,37 @@ func TestWindowAccessWindowObjects(t *testing.T) {
 		require.Equal(t, tc.detail, interceptor.Diagnostics[index].Detail)
 	}
 }
+
+func TestOldReactInternals(t *testing.T) {
+	if !isSemgrepInstalled() {
+		t.Skip("semgrep not installed, skipping test")
+		return
+	}
+	var interceptor testpassinterceptor.TestPassInterceptor
+	pass := &analysis.Pass{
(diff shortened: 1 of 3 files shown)


hugohaggmark (approved):
Looks good 🎉 could we do a similar thing looking for mutation of React props? Thank you 🙏

xnyo (approved):

jackw (approved):
LGTM! 🚀 

nit (can be ignored): depending on where it's being detected the messaging might confuse plugin developers. 🤔 

>Detected usage of React internal API '__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED'. 
This API is internal to React and should not be used directly as it may break in future React versions.

If a plugin developer directly uses it in their source code this message makes sense to me. But if it's being pulled in by a dependency at build I think it might be a bit confusing for a developer because they may not understand how to act on it. Is it possible to change the message depending on where it's found?

PR #394, opened 2025-10-08
brokenurls: do not report 429 errors in github urls

academo:
seems sending authentication token as implemented in https://github.com/grafana/plugin-validator/pull/375 will not work for github we can simply skip 429 errors for github urls

### DIFF
--- a/pkg/analysis/passes/brokenlinks/brokenlinks.go
+++ b/pkg/analysis/passes/brokenlinks/brokenlinks.go
@@ -149,6 +149,9 @@ func run(pass *analysis.Pass) (interface{}, error) {
 			defer resp.Body.Close()
 
 			if resp.StatusCode != http.StatusOK {
+				if isGitHubURL(url.url) && strings.Contains(resp.Status, "429 Too Many Requests") {
+					return
+				}
 				brokenCh <- urlstatus{url: url.url, status: resp.Status, context: url.context}
 			}
 		}(u)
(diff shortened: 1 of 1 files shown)


pkg/analysis/passes/brokenlinks/brokenlinks.go
```diff
@@ -149,6 +149,9 @@ func run(pass *analysis.Pass) (interface{}, error) {
 			defer resp.Body.Close()
 
 			if resp.StatusCode != http.StatusOK {
+				if isGitHubURL(url.url) && strings.Contains(resp.Status, "429 Too Many Requests") {
```
wbrowne, line 152: Maybe just have a check for status code on its own here since we probably don't care about any 429s in general do we?

```suggestion
				if resp.StatusCode == http.StatusTooManyRequests {
```

pkg/analysis/passes/brokenlinks/brokenlinks.go
```diff
@@ -149,7 +149,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 			defer resp.Body.Close()
 
 			if resp.StatusCode != http.StatusOK {
-				if isGitHubURL(url.url) && strings.Contains(resp.Status, "429 Too Many Requests") {
+				if resp.StatusCode == http.StatusTooManyRequests {
```
academo, line 152: it seems I merged another PR with the suggested cahnges in this PR :sight: but we should do the change suggested by Will instead.

s4kh (approved):

xnyo (approved):
LGTM!

PR #410, opened 2025-10-14
Add renovate config

academo:
Consolidates renovates PRs so we get only 1 for all dependencies instead of many (MANY!) individual ones


tolzhabayev (changes requested):
self hosted does not support `.json5`, just rename it to `.json`

tolzhabayev (approved):

xnyo (approved):

PR #419, opened 2025-10-20
osv-scanner: fix reports of vulnerabilities when multiple versions of `@grafana` packages are installed

### LINKED ISSUE #418: OSV Scanner: Vulnerabilities in Grafana Dependencies Not Filtered When Multiple Versions Present
OSV scanner reports vulnerabilities in packages that should be filtered out because they are dependencies of Grafana packages, but the filtering fails when multiple versions of a Grafana package exist in the lock file.

The `CacheGrafanaPackages()` function in `pkg/analysis/passes/osvscanner/cache-grafana-packages.go` uses `packageExists()` which performs a name-only lookup. When multiple versions of the same Grafana package exist in the parsed packages list, it always expands the **first version it finds**, not necessarily the one actually used by the project.

## Example
In `esnet-matrix-panel` with `package.json`:
```json
{
  "@grafana/data": "latest",
  "@grafana/runtime": "9.3.8"
}
```

The lock file contains two versions of `@grafana/data`:
- **9.5.1** (from `@grafana/data: "latest"`) → has `dompurify` as a direct dependency
- **9.3.8** (nested under `@grafana/runtime@9.3.8`) → does NOT have `dompurify`

The cache builder finds 9.3.8 first and expands only that version:
```
@grafana/data@9.3.8: 60 dependencies (no dompurify)
@grafana/data@9.5.1: 22 dependencies (includes dompurify)
```

Result: `dompurify@2.4.5` vulnerabilities are reported even though it comes from `@grafana/data@9.5.1` and should be filtered.

academo:
Fixes: https://github.com/grafana/plugin-validator/issues/418

The logic of the filtering remains the same but it is now expanded to find all transitive dependencies of all versions of a package inside the grafana dependencies list

e.g.
`@grafana/data@v1` brings dependencies: `a`, `b`, and `c`
`@grafana/data@v2` brings dependencies: `a`, `e`, and `f`

in the old version we'd only report `a`, `b`, and `c` as transitive dependencies of `@grafana/data` with this changes we'll report: `a`, `b`, `c`, `e` and `f`


pkg/analysis/passes/osvscanner/testdata/node/multi-version-npm/package-lock.json
andresmgot: do you really need a 25k file for this? It's impossible to know which parts are important, I would rather see a fake file with only the nested cases you are interested in testing.

andresmgot (approved):
Some comments about the maintainabilty of this code, feel free to ignore it since it's covered by tests.

PR #441, opened 2025-10-29
osvscanner: create exception mechanism for dependency@version specific packages

academo:
This creates a mechanism to except specific package@version from being reported by the osv scanner


briangann (approved):
lgtm, this will help passing e2e dependencies

PR #471, opened 2025-11-24
LLM Validator: Add SDK httpclient question

toddtreece:
This helps ensure the SDK [httpclient](https://pkg.go.dev/github.com/grafana/grafana-plugin-sdk-go/backend/httpclient#New) is used for making HTTP requests. The SDK's client includes o11y middleware, which is very helpful when debugging plugin issues.


briangann (approved):
LGTM, we should also look into automating testing this part of the validator.

toddtreece:
> LGTM, we should also look into automating testing this part of the validator.

yeah, agreed. i was looking for that and couldn't find anything

PR #474, opened 2025-12-05
Nsyncing schema

tolzhabayev:
Keeping the schemas in sync https://github.com/grafana/plugin-validator/pull/473


wbrowne (approved):
🚀

tolzhabayev:
Yep, and it's still a PR not direct push so theoretically can be reviewed to see if there is anything. I also wonder why we have two files here... But I guess we can just see what PRs look like

PR #486, opened 2026-01-13
fix: sdkusage now checking plugin's version age for warnings and errors

academo:
The previous implementation was incorrectly checking the age of the latest go-sdk release instead of the age of the version the plugin is using.

This change makes it so the validator checks the version of the grafana go sdk the plugin is using


wbrowne (approved):

xnyo (approved):
LGTM!

PR #491, opened 2026-01-21
feat: check screenshot image mimetype and extension

s4kh:
GCOM plugin publish endpoint validates image mimetype against its extension.
**Scenario**: user uploads an `screenshot2.png` but the image is actually jpeg then GCOM throws image `InvalidArgumentError`. Under the hood, it is extracting mime type from content and checking with its extension mimetype.


andresmgot (approved):
LGTM

academo (approved):

PR #497, opened 2026-02-03
runner: allow validators to access other's validators diagnostics

### LINKED ISSUE #494: Add support for validators that only run when other validators pass
## Problem

The `LLMReview` and `CodeDiff` validators use external LLM APIs which consume tokens and are expensive to run. Currently, these validators run regardless of whether other validators have already found errors in the plugin.

When a plugin has errors reported by other validators, running LLM-based validation is wasteful since the plugin will need to be fixed and resubmitted anyway.

## Proposed Solution

Add a mechanism that allows validators to declare they should only run when other validators have passed (no reported errors). This would be:

1. A property/flag on the validator itself that marks it as "run only on clean results"
2. Architecture changes to support running validators in phases or checking accumulated errors before running certain validators
3. "Errors" in this context means errors reported by validators in their reports - not warnings, not info-level messages, and not Go runtime errors

## Validators affected

- `LLMReview`
- `CodeDiff`

## Tasks

- [ ] Research current validator architecture to understand execution flow
- [ ] Design approach for conditional validator execution
- [ ] Implement the mechanism for validators to declare "run only on clean" behavior
- [ ] Update LLMReview and CodeDiff validators to use this mechanism
- [ ] Add tests

academo:
Adds infrastructure for validators to access diagnostics reported by other validators. This enables expensive validators (LLMReview, CodeDiff) to skip execution when dependencies already reported errors.

Example of Usage

 ```go
       if pass.AnalyzerHasErrors(sourcecode.Analyzer) {
           // Skip expensive LLM call
           return nil, nil
       }
       // proceed...
   }
 ```

 Notes

 - Does not modify existing validators yet - this PR only adds the infrastructure

 Closes #494

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -22,8 +22,7 @@ type Pass struct {
 	CheckParams      CheckParams
 	ResultOf         map[*Analyzer]any
 	Report           func(string, Diagnostic)
-	Diagnostics      *Diagnostics
-	AnalyzerRulesMap map[string]string
+	Diagnostics *Diagnostics
 }
 
 type CheckParams struct {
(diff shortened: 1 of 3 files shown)


xnyo (approved):
LGTM! Only one small comment

PR #496, opened 2026-01-28
runner: allow to create plugin-specific exceptions at rule level

academo:
Closes https://github.com/grafana/plugin-validator/issues/457

This PR allows to configure exceptions for specific rules for specific plugins id

This feature is a complement to the existing exceptions at a validator level but with better granularity

e.g. you want the coderules analyzer to run but only ignore specific rules for your plugin

PR #498, opened 2026-02-03
runner: fix incorrect rule exceptions logic from previous implementation

academo:
[the implementation of this functionality](https://github.com/grafana/plugin-validator/pull/496) was incorrectly using the analyzer name instead of the rule name as the key to create an exception. The tests were created using the analyzer name as the rule key and that is why they didn't catch this early on.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -57,24 +57,9 @@ func Check(
 		RootDir:     params.ArchiveDir,
 		CheckParams: params,
 		ResultOf:    make(map[*analysis.Analyzer]any),
-	}
-
-	pass.Report = func(ruleName string, d analysis.Diagnostic) {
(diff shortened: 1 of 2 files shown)


pkg/runner/runner.go
```diff
@@ -57,24 +57,9 @@ func Check(
...
-						ruleName,
-						pluginId,
-					)
-					return
-				}
-			}
-		}
-		// Collect all diagnostics for presenting at the end.
-		diagnostics[ruleName] = append(diagnostics[ruleName], d)
+		Report: func(analyzerName string, d analysis.Diagnostic) {
```
academo: this reverts to the original definition of this function before https://github.com/grafana/plugin-validator/pull/496

xnyo (approved):
LGTM!

PR #499, opened 2026-02-03
llm-review: do not run when specific validators generated errors

academo:
The LLM-review is slow and costs to run, we want to skip running it in cases where we know there are bigger errors to fix than what the LLM review could find.

I have collected a hand-picked list of validators that create strong signals of a plugin having deep problems that need fixing before an LLM can start to help.


xnyo (approved):
LGTM!

Issue #36, opened 2022-10-03
Plug-in levitate is-compatible as part of the plugin-validator report against the target grafanaDependency version


academo:
e.g.:

Validate plugins using the new topBarNavigation API to not target grafana <9.2

tolzhabayev:
added this one to the plugin submission improvements theme

academo:
After some local tries and testing I concluded that this validation is not practical.

The way grafana plugins load the grafana packages (e.g. ui, runtime, data), is taking what is provided by grafana in runtime. This runtime version doesn't necessarily match the version the plugin was compiled (and type checked) against.

Furthermore, a type conflict between packages version doesn't necessarily means the plugin won't run in runtime. E.g.: a function used to accept an string and now accepts an enum. But the runtime still handles the case for a non-defined enum key. This is something that won't compile in TS but it'll work in runtime.


This means one can't reliable detect if a plugin will be backwards compatible by checking against the types of a current version with a previous version. The check will constantly generate conflicts which even though problematic in TS won't be a problem in runtime and that can only be checked manually. This will defeat the point of an automated check.

To finish, many plugins that target older versions of grafana rely on these runtimes being compatible even if types are not, which would make it impossible to fix in the source code.

PR #500, opened 2026-02-04
Add levitate analyzer

toddtreece:
This attempts to check compatibility using levitate. I found #36, and although I agree there could be some false detections, I wonder if this falls into the same category as the LLM checks that need to be verified?

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ Run "mage gen:readme" to regenerate this section.
 -->
 | Analyzer | Description | Dependencies |
 |----------|-------------|--------------|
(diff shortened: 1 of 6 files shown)


sunker (commented):
this is an interesting idea! I could be wrong, but I think levitate's is-compatible is designed for forward compatibility checking, not backward compatibility. when checking against an older version (G10 for example), it may report APIs that were added in later versions as "Removed", which is semantically backwards as these APIs weren't removed from G10, they simply didn't exist yet. so this could be a bit confusing

PR #504, opened 2026-02-05
commit finder: improve version commit finder for codediff tool

academo:
- Fetch remote tags on shallow clone to be able to switch refs
- Fetch remote ref to be able to switch branches and tags
- Fallback into gemini-cli for commit finding


pkg/versioncommitfinder/devtest/devtest.go
```diff
@@ -0,0 +1,53 @@
+// Package main provides a manual test harness for versioncommitfinder.
```
academo: this file exist to test the gemini prompt locally. I am leaving it here for future development. 

the file itself is never imported or built in the validator.

andresmgot (approved):
LGTM, a couple of nits

PR #510, opened 2026-02-11
buildtools: add validator to check plugins are using standard build tools

academo:
New buildtools analyzer that detects plugins not built with Grafana's standard create-plugin tooling. It checks the source code for two things:

 - Frontend: Verifies the plugin uses the standard webpack configuration from @grafana/create-plugin
 - Backend: If the plugin has Go code, verifies it uses the Grafana plugin SDK build tooling


pkg/analysis/passes/buildtools/buildtools.go
```diff
@@ -0,0 +1,94 @@
...
+			"The plugin does not appear to use Grafana's standard frontend build tooling. Please use create-plugin to scaffold your plugin: https://grafana.com/developers/plugin-tools/")
+		return
+	}
+
+	b, err := os.ReadFile(webpackPath)
+	if err != nil {
+		return
+	}
+
+	if !strings.Contains(string(b), "@grafana/create-plugin") {
```
s4kh: The author can bypass by creating custom webpack.config.ts and insert `@grafana/create-plugin` comment, right? But for now this is probably the best way to test it? Would there be any other checks we can do in the module.js file?

s4kh (approved):

PR #515, opened 2026-02-16
feat: validate grafanaDep with semver

s4kh:
Validate plugin.json -> `dependencies.grafanaDependency` using SemVer constraint.
Ignores schema validation for `dependencies.grafanaDependency`.


xnyo (commented):
Some small comments but the rest LGTM to me. I don't think this will conflict with #513 as they currently do two different things, but since the analyzer in #513 is called `grafanaDependency` we can maybe move this check over there instead once merged? (before cutting a release because the analyzer name will change if we move a check from one analyzer to another)

xnyo (commented):
I went through this again considering also #513. I think we should merge this PR first, but we should make some changes before merging. My proposal is the following:

- Move the grafanaDependency check via the semver library in a separate `grafanadependency` analyzer, similar to #513. I would be fine with keeping the semver range check in the `metadatavalid` analyzer, but we need the extra checks in #513 (mostly just for our internal teams as sometimes they tend to forget the pre-release constraint -> deploy to cloud -> cause an incident because the version is "not supported". It never happens with community plugins since pretty much nobody uses on-prem "rolling" releases off main), so I think a dedicated analyzer is more suited, to keep those grafanaDependency-related checks close to each other
- Since the proposal is to move those more robust checks with the semver library in a different analyzer, I think we should remove the grafanaDependency error filtering done here: https://github.com/grafana/plugin-validator/blob/bbbfb7637ad584c6b93e05c26cafebb6f5541244/pkg/analysis/passes/metadatavalid/metadatavalid.go#L122-L127, unless it will conflict with the semver library check. I don't know if they will conflict with each other, but ideally they shouldn't: I see the regex as a first level of validation (weaker), and the semver library as a second one (more robust). If for some reason there are instances where the regex/schema check fails, but it should pass when fed into the semver library, I think we should change the regex instead of filtering out the error. If we decide to move the semver check into a dedicated analyzer, the filtering done in metadatavalid (like in this PR) will become messy because it would reference checks done on a separate `grafanadependency` analyzer. Moreover, if the user uses a config file that disables `grafanadependency` but keeps `metadatavalid`, the checks on the grafanaDependency field are skipped completely. I'd still keep a "simpler" check in metadatavalid and a proper one in a brand new analyzer

Once we merge this PR, I will rework #513 also considering your previous feedback (this PR is in a better state than #513, we're better off unblocking this one first)

What are your thoughts on this? @s4kh @academo

xnyo (approved):
Just one small comment, the rest LGTM! Great job! 🙌

PR #518, opened 2026-02-17
feat: mcp server for validate plugin

s4kh:
Exposes plugin-validator CLI as MCP server. Tool: `validate_plugin`
1. Separate release for mcpserver - published to npm
3. MCP server uses npx to run the latest validator CLI
4. Includes configs for vscode extension, claude code, openAI codex configs in the README


pkg/cmd/mcpserver/main.go
```diff
@@ -0,0 +1,62 @@
...
+		Description: "Validates a Grafana plugin against publishing requirements. Checks metadata, security, structure, and best practices. Returns detailed errors and warnings with actionable fix suggestions.",
+	}, ValidatePlugin)
+	if err := server.Run(context.Background(), &mcp.StdioTransport{}); err != nil {
+		return fmt.Errorf("failed to run server: %w", err)
+	}
+	return nil
+}
+
+func main() {
+	if err := run(); err != nil {
```
s4kh: https://grafana.com/blog/how-i-write-http-services-in-go-after-13-years/ -> `func main()` only calls `run()`

academo (commented):
I would like you to reconsider the approach for the mcp server

the validator already supports outputting as json, could you  invoke the validator binary here and parse the output instead?

here are my two concerns with this approach:

1) single source of truth: now we are effectively creating two different software that embed the same engine. I much rather prefer we keep a single entry point (the validator cli binary) and make MCP just another user, like for example github actions is an user of the validator, and ci-workflows, and the community pipeline. 

We don't need to create separate binaries for those cases because we have a common serialized output any user can read.

2) updates: this binary will be out of date very quickly. instead we can have this wrapper checking the latest version (once a day?), downloading it and executing it to get the json output. 

I don't see a mechanism to update mcp servers, they need to auto-update. People that use the validator mainly use it via docker or npx that always pulls the latest version (except NPM when it wants to be a pain)


my bottom line: we can make this wrapper simply call `npx` or `docker` or `yarn dlx` or similar (support several fallbacks) and parse the json output.

pkg/cmd/mcpserver/main.go
```diff
@@ -0,0 +1,264 @@
...
+	Title    string `json:"Title"`
+	Detail   string `json:"Detail"`
+	Name     string `json:"Name"`
+}
+
+// Diagnostics is a map of category name to list of diagnostics
+type Diagnostics map[string][]Diagnostic
+
+// Severity constants
+const (
```
academo: can we import the types from the existing definitions instead of re-defining them here?

pkg/utils/utils.go
```diff
@@ -65,3 +66,25 @@ func GetPluginMetadata(archiveDir string) (*metadata.Metadata, error) {
 	}
 	return &pluginJson, nil
 }
+
+// HasProperArchiveStructure checks if the archive has the proper structure:
+// single top-level directory containing plugin.json
+func HasProperArchiveStructure(archiveDir string) bool {
```
academo: is this function used anywhere?

academo (commented):
It is going towards the right place, my main points:

* create an npm package and docker image to distribute it
* update the instructions to use such package or image
* do not redefine types that are already available

I am not truly convinced the "summary" is adding any value. it is the LLM that will get this input and they can already figure out what is what from the validator output as is

.github/workflows/do-release-mcp.yml
```diff
@@ -0,0 +1,69 @@
...
+          git config user.email '144369747+grafana-plugins-platform-bot[bot]@users.noreply.github.com'
+
+      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+        with:
+          node-version: "24"
+
+      - name: Bump MCP version and create tag
+        run: |
+          cd mcp-package
+          npm version ${INPUT_VERSION} --no-git-tag-version
```
s4kh: `--no-git-tag-version` so it does not auto commit and tag `0.0.1`. We need to `mcp/v` prefix.

pkg/cmd/mcpserver/main.go
academo: importing the types from another go package will not increase the size because after compilation it'll strip out all that it doesn't use. You need to compile with the correct optimization flag.

.github/workflows/release-mcp.yml
```diff
@@ -0,0 +1,101 @@
...
+          persist-credentials: false
+
+      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+        with:
+          node-version: "24"
+          registry-url: "https://registry.npmjs.org"
+      - run: cd mcp-package && npm install
+      - run: cd mcp-package && npm publish
+
+  release-mcp-to-dockerhub:
```
academo: since the running docker inside docker is so problematic,. let's skip the docker release?

.github/workflows/release-mcp.yml
```diff
@@ -0,0 +1,101 @@
...
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - run: git fetch --force --tags
```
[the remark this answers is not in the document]
s4kh: Yeah it came in from release.yml -> which was a requirement for old gorealeaser for generating the correct changelog.

academo (approved):

PR #520, opened 2026-02-17
llmclient: add agentic client for code reviews

academo:
this PR adds a basic agent client that can use openai/anthropic/google apis to answer questions about code.

The agent is barebones and it is not yet used anywhere. The idea is introduce the client here and then move code-diff to use it directly instead of the gemini-cli

## AgenticClient

The agentic client runs a tool-calling loop where the LLM can explore code using read-only tools, then submits structured answers via `submit_answer`.

**Tools**: `list_directory`, `read_file`, `grep`, `git` (allowlisted subcommands), `submit_answer`

**Providers**: Google (Gemini), Anthropic (Claude), OpenAI

```mermaid
sequenceDiagram
    participant Caller
    participant AgenticClient
    participant LLM
    participant Tools

    Caller->>AgenticClient: CallLLM(prompt, repoPath, opts)
    AgenticClient->>LLM: system prompt + user prompt + tool definitions

    loop until submit_answer or 100 tool calls
        LLM-->>AgenticClient: tool call(s)
        alt submit_answer
            AgenticClient->>AgenticClient: collect answer
            AgenticClient-->>LLM: "Answer recorded"
        else read_file / list_directory / grep / git
            AgenticClient->>Tools: execute tool (sandboxed to repoPath)
            Tools-->>AgenticClient: result
            AgenticClient-->>LLM: tool result
        end
    end

    Note over LLM,AgenticClient: LLM sends message with no tool calls → done
    AgenticClient-->>Caller: []AnswerSchema
```

### DIFF
--- a/pkg/llmclient/agentic_client.go
+++ b/pkg/llmclient/agentic_client.go
@@ -1,0 +1,320 @@
+package llmclient
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/tmc/langchaingo/llms"
+	"github.com/tmc/langchaingo/llms/anthropic"
+	"github.com/tmc/langchaingo/llms/googleai"
+	"github.com/tmc/langchaingo/llms/openai"
+)
+
+const (
+	maxToolCalls            = 100
+	maxLLMRetries           = 3
+	maxConsecutiveNoTools   = 5
+	retryDelay              = 2 * time.Second
+)
+
+// AnswerSchema represents the structured response from the agentic client
+type AnswerSchema struct {
+	Answer      string   `json:"answer"`
+	ShortAnswer bool     `json:"short_answer"`
+	Files       []string `json:"files,omitempty"`
+	CodeSnippet string   `json:"code_snippet,omitempty"`
+}
+
+// AgenticCallOptions contains configuration for the agentic LLM call
+type AgenticCallOptions struct {
+	Model    string // e.g. "gemini-2.0-flash"
+	Provider string // "google", "anthropic", "openai"
+	APIKey   string
+}
+
+// AgenticClient is an interface for agentic LLM interactions
+type AgenticClient interface {
+	CallLLM(ctx context.Context, prompt, repositoryPath string, opts *AgenticCallOptions) ([]AnswerSchema, error)
+}
+
+// agenticClientImpl implements AgenticClient
+type agenticClientImpl struct{}
+
+// NewAgenticClient creates a new AgenticClient
+func NewAgenticClient() AgenticClient {
+	return &agenticClientImpl{}
+}
+
+// CallLLM executes an agentic loop with tools to answer questions about code.
+// The prompt may contain multiple questions, in which case the agent will call
+// submit_answer multiple times. All answers are collected and returned.
+func (c *agenticClientImpl) CallLLM(ctx context.Context, prompt, repositoryPath string, opts *AgenticCallOptions) ([]AnswerSchema, error) {
+	if opts == nil {
+		return nil, fmt.Errorf("options are required")
+	}
+
+	if opts.APIKey == "" {
+		return nil, fmt.Errorf("API key is required")
+	}
+
+	if opts.Model == "" {
+		return nil, fmt.Errorf("model is required")
+	}
+
+	if opts.Provider == "" {
+		return nil, fmt.Errorf("provider is required")
+	}
+
+	// Initialize LLM based on provider
+	llm, err := initLLM(ctx, opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize LLM: %w", err)
+	}
+
+	// Build tools
+	tools := buildAgenticTools()
+
+	// Create tool executor
+	executor := newToolExecutor(repositoryPath)
+
+	// System prompt
+	systemPrompt := `You are a code analysis assistant. You have tools to explore code in a repository.
+
+AVAILABLE TOOLS:
+- list_directory: List files at a path. Use "." for root.
+- read_file: Read a file's contents. This is your primary tool for understanding code.
+- grep: Search for a pattern across files.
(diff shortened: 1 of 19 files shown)


pkg/llmclient/testdata/fs_access/storage/disk.go
```diff
@@ -0,0 +1,66 @@
...
+	"os"
+	"path/filepath"
+)
+
+type DiskStore struct {
+	baseDir string
+}
+
+func NewDiskStore(baseDir string) (*DiskStore, error) {
+	if err := os.MkdirAll(baseDir, 0755); err != nil {
```
s4kh: Should we use a temp directory here?

pkg/llmclient/agentic_client.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	// Build tools
+	tools := buildAgenticTools()
+
+	// Create tool executor
+	executor := newToolExecutor(repositoryPath)
+
+	// System prompt
+	systemPrompt := `You are a code analysis assistant. You have tools to explore code in a repository.
```
s4kh: I am wondering if we can use something from here https://github.com/anthropics/claude-code/blob/main/plugins/code-review/commands/code-review.md

pkg/llmclient/agentic_tools.go
```diff
@@ -0,0 +1,339 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/tmc/langchaingo/llms"
+)
+
+const maxFileSize = 500 * 1024 // 500KB
+
+// Git subcommands that are allowed
+var allowedGitSubcommands = map[string]bool{
```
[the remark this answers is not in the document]
academo: exactly.

s4kh (approved):

pkg/llmclient/agentic_client.go
```diff
@@ -0,0 +1,320 @@
...
+
+AVAILABLE TOOLS:
+- list_directory: List files at a path. Use "." for root.
+- read_file: Read a file's contents. This is your primary tool for understanding code.
+- grep: Search for a pattern across files.
+- git: Run read-only git commands (log, show, diff, status, etc.)
+- submit_answer: Submit your final answer.
```
xnyo: This looks a bit off to me: do we have to specify this if we already define a description for each tool in `buildAgenticTools`? Doesn't langchaingo put those in the prompt already?

pkg/llmclient/agentic_client.go
```diff
@@ -0,0 +1,320 @@
...
+		if choice.Content != "" {
+			debugLog("AgenticClient: AI message: %s", truncateString(choice.Content, 200))
+		}
+
+		// If no tool calls, check if we have answers
+		if len(choice.ToolCalls) == 0 {
+			debugLog("AgenticClient: no tool calls in response")
+
+			// If we have collected answers, the agent is done
+			if len(answers) > 0 {
```
xnyo: What if the agent submits some answer(s), replies but wants to read more files before giving the final answer? In this case we'll return as soon as we have one answer. How do we know that the agent has finished checking all files? Maybe we can add another "done" tool, or we can prompt it to check all files and submit an answer for each file (will cost more tokens)

pkg/llmclient/agentic_tools.go
```diff
@@ -0,0 +1,339 @@
...
+	repoPath string
+}
+
+// newToolExecutor creates a new tool executor for the given repository path
+func newToolExecutor(repoPath string) *toolExecutor {
+	return &toolExecutor{repoPath: repoPath}
+}
+
+// execute runs a tool and returns the result as a string
+func (e *toolExecutor) execute(toolName, argsJSON string) string {
```
xnyo: Return a `string, error` here rather than using the string as an error, then the caller can decide how to pass the error back to the LLM

pkg/llmclient/agentic_tools.go
```diff
@@ -0,0 +1,339 @@
...
+func (e *toolExecutor) readFile(args map[string]interface{}) string {
+	path, ok := args["path"].(string)
+	if !ok {
+		return "Error: path is required"
+	}
+
+	fullPath, errMsg := e.validatePath(path)
+	if errMsg != "" {
+		return errMsg
+	}
```
xnyo: Make each function return a proper `error` value rather than relying on strings:


```suggestion
	fullPath, err := e.validatePath(path)
	if err != "" {
		return "", fmt.Errorf("validate path: %w", err)
	}
```

pkg/llmclient/agentic_tools.go
```diff
@@ -0,0 +1,339 @@
...
+	if info.Size() > maxFileSize {
+		return fmt.Sprintf("Error: file is too large (%d bytes, limit is %d bytes). Try reading a smaller file or use grep to find specific content.", info.Size(), maxFileSize)
+	}
+
+	content, err := os.ReadFile(fullPath)
+	if err != nil {
+		return fmt.Sprintf("Error reading file: %v", err)
+	}
+
+	return string(content)
```
xnyo: This can be a problem if the file a binary one. The file size check above kind of protects against that, but I think we should be more strict:


```suggestion
	r := string(content)
	if !utf8.ValidString(r) {
		return "", errors.New("file is not a text file")
	}
	return string(content)
```

xnyo (commented):
Awesome work! I have some comments but I really like this idea, much much better than the Gemini CLI approach we currently have 🥇 🙌

pkg/llmclient/agentic_client.go
academo: the agent will keep calling tools on each turn. when an agent stops calling tools is when we consider the loop finished and only if we don't have answers we return

because the questions come in a prompt we don't know many answer we need. the prompt might be asking a single question or 3.

xnyo (approved):
LGTM, awesome work! 🙌 🚀

PR #533, opened 2026-03-09
agentic client: reset context over 100k tokens

academo:
If we are about to start a new question loop we should reset the context if the token count is above 100k

above 100k (50% context of most context) the agents start performing poorly plus this avoids possibly the model running out of context if there are many questions coming in.


xnyo (approved):
LGTM! Only one small nit

PR #534, opened 2026-03-09
codediff: use new agentic client instead of gemini-cli

academo:
It uses the recently introduced agentic client instead of the gemini-cli to perform the code-diff


mckn (approved):
LGTM! Great work on this

PR #535, opened 2026-03-09
ref: report code diff version report as info

s4kh:
Fixes https://github.com/grafana/community-pipeline/issues/141
Report code diff versions as info


mckn (approved):
LGTM!

academo (changes requested):
We don't want to show this info to users, this is intentional a "suspected" because we don't want to show users potential false positives.

I am going to work in this report in the coming weeks.

PR #537, opened 2026-03-10
feat: disable some analyzers for Grafana Labs plugins

xnyo:
Provides a better default configuration for Grafana Labs (internally developed) plugins now that we are testing running plugin-validator as part of https://github.com/grafana/plugin-ci-workflows.

This PR skips the following analyzers are for Grafana Labs plugins:

- discoverability
- license
- org (not really needed but saves us a GCOM request)
- sponsorshiplink

Other analyzers that are disabled via other methods:
- codediff (LLM) -> disabled via env var (requires gemini API key)
- llmreview -> disabled via env var (requires gemini API key)
- virusscan -> disabled via env var (too expensive/slow in CI, and not that useful for internal plugins)
- safelinks: disabled via env var (requires webrisk API key)

Other analyzers that were considered but not turned off:

- pluginname -> check for plugin slug format skipped if plugin is already PUBLISHED: enforces the check only for new plugins, which is good also for `grafana-*` plugins as a sanity check
- trackingscripts -> seems useful and precise enough and not expensive to run

### DIFF
--- a/Magefile.go
+++ b/Magefile.go
@@ -55,6 +55,8 @@ func buildDockerImage() error {
 	return sh.RunV(
 		"docker",
 		"build",
+		"--platform",
+		"linux/amd64",
 		"--no-cache",
 		"--pull",
 		"-t",
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,8 @@
+package config
+
+import (
+	_ "embed"
+)
+
+//go:embed grafana.yaml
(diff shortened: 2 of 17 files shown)


academo (changes requested):
we have a config file for theses cases.

I much prefer we even embed a config file only for grafana-specific plugins and load it if we detect it (and no config file was passed) than to hardcode this logic into every validator. 

can you look into that option?

PR #546, opened 2026-03-20
Add semgrep test to avoid calling /api/ds/query directly

andresmgot:
This endpoint will eventually go away once `/apis/...` is the only endpoint supported.


toddtreece (approved):
good idea!

xnyo (approved):
LGTM!

PR #549, opened 2026-03-25
Add stale backend build check

toddtreece:
Adds a new gobuildinfo analyzer that validates embedded Go build metadata in backend plugin binaries. The checks catch common issues that indicate a binary was not built correctly for submission:
                                                                                                                                                                                                                     
  - binary-no-build-info: binary is missing Go build metadata (e.g. was stripped)                                                                                                                                    
  - binary-dirty-build: binary was built from an uncommitted working tree (vcs.modified=true)                                                                                                                        
  - binary-cgo-enabled: binary was built with CGO_ENABLED=1, which is unsupported                                                                                                                                    
  - binary-plugin-id-mismatch: plugin ID embedded via -ldflags doesn't match plugin.json                                                                                                                             
  - binary-build-info-json-plugin-id-mismatch: SDK buildInfoJSON plugin ID doesn't match plugin.json                                                                                                                 
  - binary-build-info-json-version-mismatch: SDK buildInfoJSON version doesn't match plugin.json
  - binary-dep-not-in-gomod: a compiled-in dependency is absent from the submitted go.mod                                                                                                                            
  - binary-dep-gomod-version-mismatch: a compiled-in dependency is at a different version than declared in go.mod

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ Run "mage gen:readme" to regenerate this section.
 | Code Rules / `code-rules` | Checks for forbidden access to environment variables, file system or use of syscall module. | [semgrep](https://github.com/returntocorp/semgrep), `sourceCodeUri` |
 | Developer Jargon / `jargon` | Generally discourages use of code jargon in the documentation. | None |
 | Discoverability / `discoverability` | Warns about missing keywords and description that are used for plugin indexing in the catalog. | None |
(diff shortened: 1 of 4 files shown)


academo (approved):

academo:
@toddtreece do you want to follow up on this?

PR #550, opened 2026-03-26
fix: make LLM review resilient to individual question failures

academo:
The LLM review was silently dropping all results when any single question failed. Two issues:

- `AskLLMAboutCode` had a retry loop wrapping `CallLLM`, which itself already retries internally via `callLLMWithRetry`  so a failing question would fire up to 9 API calls before giving up, then discard all previously successful answers
- The 20s per-call timeout was too short for large plugin codebases, causing frequent `context deadline exceeded` errors on Anthropic

Now failed questions are skipped and processing continues, the redundant outer retry is removed, and the timeout is bumped to 90s. Also adds an OK report when LLM review completes clean.

### DIFF
--- a/pkg/analysis/passes/llmreview/llmreview.go
+++ b/pkg/analysis/passes/llmreview/llmreview.go
@@ -34,6 +34,10 @@ var (
 		Name:     "llm-review-skipped",
 		Severity: analysis.SuspectedProblem,
 	}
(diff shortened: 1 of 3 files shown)


tolzhabayev (approved):

xnyo (approved):
LGTM!

PR #553, opened 2026-03-30
llmreview: optimize prompts for concise LLM output

academo:
LLM review answers were extremely verbose. This tunes the system prompt, tool descriptions, and question text to get concise, investigative-only responses.


leventebalogh (approved):
LGTM 👍

PR #582, opened 2026-05-19
fix: code-rules per-rule exceptions and severity config not applied

academo:
Rules like code-rules-no-direct-css-imports were allocated fresh inside run() on each semgrep result, so initAnalyzers() could never mark them disabled. Pre-registering all semgrep rules at init() time fixes this.

### DIFF
--- a/pkg/analysis/passes/coderules/coderules.go
+++ b/pkg/analysis/passes/coderules/coderules.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/plugin-validator/pkg/analysis"
 	"github.com/grafana/plugin-validator/pkg/analysis/passes/sourcecode"
 	"github.com/grafana/plugin-validator/pkg/logme"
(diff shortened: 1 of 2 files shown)


pkg/analysis/passes/coderules/coderules.go
```diff
@@ -48,6 +54,35 @@ var Analyzer = &analysis.Analyzer{
...
+	Severity string `yaml:"severity"`
+}
+
+type semgrepRulesFile struct {
+	Rules []semgrepRuleEntry `yaml:"rules"`
+}
+
+func init() {
+	var rulesFile semgrepRulesFile
+	if err := yaml.Unmarshal([]byte(semgrepRules), &rulesFile); err != nil {
```
academo, line 68: file comes from embed `semgrep-rules.yaml`

andresmgot (approved):

PR #585, opened 2026-05-19
fix: accept lowercase magefile.go in backend build tooling check

academo:
The check was doing an exact case match on `Magefile.go`, failing on `magefile.go`. Both are valid - mage itself accepts either. Switches to a glob pattern `[Mm]agefile.go` to handle both.


oshirohugo (approved):

mckn (approved):
LGTM!

PR #593, opened 2026-05-26
reactcompat: run in CI and make output reproducible

academo:
Docker builder was missing nodejs/npm, so reactcompat skipped in CI but ran locally, and react-detect emitted machine-specific absolute paths.

Adds pinned nodejs 24.14.1-r0 + npm to both Dockerfile stages (bumps Alpine to 3.23 for Node 24), sets the command cwd to the archive dir so reported paths can be stripped to relative, and updates the integration test expectations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ ARG GOLANGCI_LINT_VERSION=v2.5.0
 ARG GOSEC_VERSION=v2.22.8
 ARG SEMGREP_VERSION=1.84.1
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ ARG GOLANGCI_LINT_VERSION=v2.5.0
 ARG GOSEC_VERSION=v2.22.8
```
academo, line 2: make sure nodejs is installed so we can run the reactcompact tests in CI

andresmgot (approved):

PR #611, opened 2026-06-09
Fix slack notification about staged npm packagec

Ukochka:
Update Slack action to v3.0.1


.github/workflows/test-slack.yml
xnyo, line 1: Should this file be removed?

xnyo (commented):
LGTM but one question before approving:

.github/workflows/test-slack.yml
Ukochka: Yes, it is removed! I was testing the notification

academo (approved):

PR #614, opened 2026-06-12
govulncheck: make output less verbose

toddtreece:
Example output:
```
Update Go toolchain to 1.26.4 or later (GO-2026-4971, GO-2026-4976, GO-2026-4977, GO-2026-4980, GO-2026-4981, GO-2026-4982, GO-2026-4986, GO-2026-5037, GO-2026-5038, GO-2026-5039)

Update the following dependencies:
• golang.org/x/net v0.55.0 (GO-2026-4918, GO-2026-5025, GO-2026-5026, GO-2026-5027, GO-2026-5028, GO-2026-5029, GO-2026-5030)
• golang.org/x/sys v0.44.0 (GO-2026-5024)
```

### DIFF
--- a/pkg/analysis/passes/govulncheck/govulncheck.go
+++ b/pkg/analysis/passes/govulncheck/govulncheck.go
@@ -43,7 +43,6 @@ var Analyzer = &analysis.Analyzer{
 
 type vulnInfo struct {
 	id           string
-	summary      string
 	module       string
 	fixedVersion string
 }
@@ -123,7 +122,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 				continue
 			}
 			for id, info := range vulns {
-				if sourceFindings[id] == nil {
+				if sourceFindings[id] == nil || semver.Compare(info.fixedVersion, sourceFindings[id].fixedVersion) > 0 {
 					sourceFindings[id] = info
 				}
 			}
@@ -228,7 +227,6 @@ func runGovulncheckJSON(govulncheckBin, dir, target string, args ...string) ([]b
 // symbol is reachable from user code, not merely present in a transitive dep).
 func parseCalledFindings(r io.Reader) (map[string]*vulnInfo, error) {
 	dec := json.NewDecoder(r)
-	summaries := make(map[string]string)
 	called := make(map[string]*vulnInfo)
 	for {
 		var msg Message
@@ -237,9 +235,6 @@ func parseCalledFindings(r io.Reader) (map[string]*vulnInfo, error) {
 				break
 			}
 			return nil, err
-		}
-		if msg.OSV != nil && msg.OSV.ID != "" {
-			summaries[msg.OSV.ID] = msg.OSV.Summary
 		}
 		if msg.Finding == nil || msg.Finding.OSV == "" {
 			continue
@@ -257,15 +252,11 @@ func parseCalledFindings(r io.Reader) (map[string]*vulnInfo, error) {
 			}
 		}
 	}
-	for id, info := range called {
-		info.summary = summaries[id]
-	}
 	return called, nil
 }
 
 func parseAllFindings(r io.Reader) (map[string]*vulnInfo, error) {
 	dec := json.NewDecoder(r)
-	summaries := make(map[string]string)
 	found := make(map[string]*vulnInfo)
 	for {
 		var msg Message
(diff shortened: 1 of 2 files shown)


s4kh (dismissed):

pkg/analysis/passes/govulncheck/govulncheck.go
```diff
@@ -228,19 +238,35 @@ func parseCalledFindings(r io.Reader) (map[string]struct{}, error) {
...
-			called[msg.Finding.OSV] = struct{}{}
+			id := msg.Finding.OSV
+			if called[id] == nil {
+				called[id] = &vulnInfo{id: id}
+			}
+			if semver.Compare(msg.Finding.FixedVersion, called[id].fixedVersion) > 0 {
+				called[id].fixedVersion = msg.Finding.FixedVersion
+			}
+			if called[id].module == "" {
+				called[id].module = firstModule(msg.Finding.Trace)
```
mckn, line 251: There might be a bug in the "firstModule" function according to this:

According to the govulncheck JSON spec, Finding.Trace for source-mode is a call stack where trace[0] is the user's entry point and subsequent frames descend into the vulnerable dependency. So firstModule(msg.Finding.Trace) returns the user's own module, not the vulnerable dependency.

For example, for a plugin calling into golang.org/x/net, the trace is:
```go
trace[0]: {module: "github.com/user/myplugin", function: "main"}
trace[1]: {module: "golang.org/x/net", function: "vulnerableFunc"}

```
firstModule returns "github.com/user/myplugin", and the output becomes:
Update the following dependencies:
• github.com/user/myplugin v1.2.3 (GO-2024-AAAA)   ← wrong

The fix should look at the last non-empty module in the trace (or specifically the frame that has Position == nil, which is the vulnerable symbol frame). The test data in sampleNDJSON happens to use the same module for all frames, so the unit test passes but doesn't catch this.

pkg/analysis/passes/govulncheck/govulncheck.go
```diff
@@ -39,6 +41,13 @@ var Analyzer = &analysis.Analyzer{
 	},
 }
 
+type vulnInfo struct {
+	id           string
+	summary      string
```
mckn: Are we ever using this field? I can see that we set it but I can't find any usages of reading it except in tests.

pkg/analysis/passes/govulncheck/govulncheck.go
```diff
@@ -113,15 +122,17 @@ func run(pass *analysis.Pass) (interface{}, error) {
 				)
 				continue
 			}
-			for id := range osvIDs {
-				sourceFindings[id] = struct{}{}
+			for id, info := range vulns {
+				if sourceFindings[id] == nil {
+					sourceFindings[id] = info
+				}
```
mckn: Should we set this not only based on if it is nil but also set it when `semver.Compare` is greater than the previous set one so we always use the newest.

mckn (commented):
Looks great! Added some small nits to consider.

pkg/analysis/passes/govulncheck/govulncheck.go
toddtreece: good catch on the test, the fixture used the same module on every frame, so it didn't actually check which end firstModule reads from. updated in https://github.com/grafana/plugin-validator/pull/614/commits/bee6cb58b09d997dffe9b8aa4de74b09c1491ef0

the `Finding.Trace` doc in ([govulncheck.go#L154-L156](https://github.com/golang/vuln/blob/v1.1.4/internal/govulncheck/govulncheck.go#L154-L156)) states:

> Frames are sorted starting from the imported vulnerable symbol until the entry point. The first frame in Frames should match Symbol.

it seems like `trace[0]` is the vulnerable dependency and the entry point is last. govulncheck's own renderer reads `Trace[0].Module` as the affected module ([text.go#L279](https://github.com/golang/vuln/blob/v1.1.4/internal/scan/text.go#L279)).

pkg/analysis/passes/govulncheck/govulncheck.go
toddtreece: thanks! removed in https://github.com/grafana/plugin-validator/pull/614/commits/8cf02f5dfbdd364122c87570eddda4ef5e3f244c

pkg/analysis/passes/govulncheck/govulncheck.go
toddtreece: good call. updated in https://github.com/grafana/plugin-validator/pull/614/commits/8cf02f5dfbdd364122c87570eddda4ef5e3f244c

academo (approved):

PR #618, opened 2026-06-19
chore: bump osv-scanner to v2.4.0 and go to 1.26.4

tolzhabayev:
Bumps `github.com/google/osv-scanner/v2` from v2.3.8 to [v2.4.0](https://github.com/google/osv-scanner/releases/tag/v2.4.0).

v2.4.0 requires Go >= 1.26.4, so I bumped the `go` directive and the Dockerfile builder image to 1.26.4 (the latest stable release — 1.27 is still an RC). The CI workflows read the version from go.mod via `go-version-file`, so they pick this up automatically.

Also dropped the `runtime-spec` v1.2.1 replace directive. It was there to keep containerd v1.7.x compilable, but v2.4.0 moves osv-scanner onto containerd v2, so the pin is no longer needed and the build is clean without it.

Build and the osv-scanner pass tests are green. The only failing tests are the pre-existing `reactcompat` integration tests, which fail identically on main (environment-dependent `react-detect` output).


Dockerfile
```diff
@@ -54,13 +54,13 @@ RUN mage -v build:lint
 
 RUN mage -v build:ci
 
-FROM alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+FROM golang:1.26.4-alpine3.24@sha256:3ad57304ad93bbec8548a0437ad9e06a455660655d9af011d58b993f6f615648
```
tolzhabayev: there you go @academo

Dockerfile
```diff
@@ -54,13 +54,13 @@ RUN mage -v build:lint
 
 RUN mage -v build:ci
 
-FROM alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+FROM golang:1.26-alpine@sha256:3ad57304ad93bbec8548a0437ad9e06a455660655d9af011d58b993f6f615648
```
tolzhabayev, line 62: @academo switched to floating tag

academo: it's still pinned to the hash so it makes no difference.

academo (dismissed):

academo (approved):

PR #642, opened 2026-07-14
feat(backendbinary): require a linux/amd64 backend binary

academo:
When a plugin declares a backend and matching binaries are present, the backendbinary pass now requires at least one to target linux/amd64. Platform is read from the embedded Go build info (GOOS/GOARCH), the same approach govulncheck uses, so it works on stripped production builds and does not trust file names.

- New `backend-binary-not-linux-amd64` rule (error) when no linux/amd64 binary is found
- Tests now build real cross-compiled binaries instead of empty placeholder files

Also fixes a pre-existing map-order bug where a frontend plugin.json with no executable could abort the whole pass and skip validating a nested backend.

### DIFF
--- a/pkg/analysis/passes/backendbinary/backendbinary.go
+++ b/pkg/analysis/passes/backendbinary/backendbinary.go
@@ -1,6 +1,7 @@
 package backendbinary
 
 import (
+	"debug/buildinfo"
 	"fmt"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


pkg/analysis/passes/backendbinary/backendbinary.go
```diff
@@ -99,7 +105,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 		// no executable in plugin.json skipping other checks
 		if data.Executable == "" {
-			return nil, nil
+			continue
```
academo, line 108: fixes a bug, we were potentially skipping nested backend binaries.

tolzhabayev (approved):

PR #858, opened 2026-09-11
fix: log scanner memory and report incomplete scans

tolzhabayev:
A killed scanner or partial report could leave no failure diagnostic. Semgrep, ClamAV and gosec now report `scan-incomplete` when execution fails. Partial Semgrep scans, malformed reports and unexpected analyzer errors also produce that diagnostic. Existing findings are preserved, and disabled analyzers stay disabled.

Analyzer and scanner start/end logs include duration, exit code and process peak RSS on Linux/macOS. The last start event identifies the active scanner if an OOM kills the container. RSS covers the process rather than the combined memory of all its children. The logs exclude arguments and source contents.

The full `mage test:default` suite and `mage build:lint` passed with a Go 1.27-compatible linter. Tests cover killed scanners, incomplete reports and valid findings. The tests also caught a JSX fixture saved as `.ts`, which is now `.tsx`.

Deploy [the review-service change](https://github.com/grafana/plugin-review-service/pull/246) first, then release this validator and update the service's image pin. Keep the 8 GiB worker limit while collecting measurements. Memory tuning follows once we identify the scanner responsible.


README.md
```diff
@@ -415,6 +415,22 @@ global:
   reportAll: true
 ```
 
+### Scanner diagnostics
```
academo: this readme addition is unnecessary.

pkg/analysis/passes/coderules/coderules.go
```diff
@@ -128,16 +130,29 @@ func run(pass *analysis.Pass) (any, error) {
...
 	err = json.Unmarshal(out, &semgrepResults)
 	if err != nil {
+		pass.ReportIncomplete("Could not decode Semgrep output: " + err.Error())
 		return nil, nil
 	}
+	if semgrepResults.Results == nil {
+		pass.ReportIncomplete("Semgrep output is missing scan results")
+		return nil, nil
+	}
+	if len(semgrepResults.Errors) > 0 {
```
academo, line 149: I am not sure I get this part of the logic. are we reporting incomplete scans when errors are found? we'll report these errors down the line.

pkg/analysis/passes/gosec/gosec.go
```diff
@@ -71,28 +73,26 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		var exitErr *exec.ExitError
+		if !errors.As(err, &exitErr) || exitErr.ExitCode() != 1 {
+			pass.ReportIncomplete("gosec execution failed: " + err.Error())
+			return nil, nil
 		}
 	}
 
 	if len(goSecOutput) == 0 {
-		logme.Debugln("gosec output is empty, skipping gosec report")
+		pass.ReportIncomplete("gosec returned an empty report")
```
academo, line 91: are we fully confident that an empty gosec report means is incomplete?

pkg/analysis/passes/gosec/gosec.go
```diff
@@ -71,28 +73,26 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 	}
 
 	var goSectResults Result
 	err = json.Unmarshal(goSecOutput, &goSectResults)
 	if err != nil {
-		fmt.Println("Error running gosec", err)
-		logme.Errorln("Error unmarshalling gosec output", "error", err)
-		// breaking the validator to notify the user that the gosec output is not as expected
-		return nil, err
+		pass.ReportIncomplete("Could not decode gosec output: " + err.Error())
```
academo, line 99: please keep the logme.Errorln line here.

pkg/analysis/passes/virusscan/virusscan.go
```diff
@@ -81,7 +83,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	err = runClamavScan(clamavBin, archiveDir, "archive", pass)
 	if err != nil {
-		logme.Debugln("clamav failed, skipping virus scan", err)
+		pass.ReportIncomplete(err.Error())
```
academo: you are missing the verbose first part of the report incomplete like you did for other validators.

pkg/analysis/passes/virusscan/virusscan.go
```diff
@@ -105,18 +107,20 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	clamavOutput, err := scanprocess.CombinedOutput(clamavCommand)
 
-	// clamav exits 1 if it finds issues. Only failing if the output is empty
-	if err != nil && len(clamavOutput) == 0 {
-		logme.Debugln("clamav failed, skipping virus scan", err)
-		return nil
+	// Exit 1 means malware was found. Other failures must never become a clean scan.
+	var exitErr *exec.ExitError
+	if err != nil && (!errors.As(err, &exitErr) || exitErr.ExitCode() != 1) {
+		return fmt.Errorf("ClamAV %s scan failed: %w", entityName, err)
```
academo: do not turn this into an error (or others down below). this literallty breaks the validator process and it won't even report this found virus

README.md
tolzhabayev: removed

pkg/analysis/passes/coderules/incomplete_test.go
```diff
@@ -0,0 +1,35 @@
...
+		{"invalid report", `not JSON`, ""},
+		{"missing results", `{}`, ""},
+		{"partial scan", `{"results":[],"errors":[{"type":"OutOfMemory","message":"memory limit reached"}]}`, ""},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			directory := t.TempDir()
+			script := "#!/bin/sh\nprintf '%s\\n' '" + tc.output + "'\n" + tc.command + "\n"
+			require.NoError(t, os.WriteFile(filepath.Join(directory, "semgrep"), []byte(script), 0755))
+			t.Setenv("PATH", directory)
+			var diagnostics []analysis.Diagnostic
```
academo, line 26: most of the logic here and all the other tests are doing what the testinterceptor already does https://github.com/grafana/plugin-validator/blob/bc4bb35f01447cca2ba63265710e7bc93a604ff1/pkg/testpassinterceptor/testpassinterceptor.go#L9


can you update all the tests to use the existing pattern?

pkg/analysis/analysis.go
```diff
@@ -17,6 +17,7 @@ var (
 )
 
 type Pass struct {
+	Analyzer     *Analyzer
```
academo, line 20: this type change doesn't look necessary (only for tests?)

pkg/analysis/passes/virusscan/virusscan.go
```diff
@@ -79,11 +81,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	logme.DebugFln("Will run clamav on %s", archiveDir)
 
-	err = runClamavScan(clamavBin, archiveDir, "archive", pass)
-	if err != nil {
-		logme.Debugln("clamav failed, skipping virus scan", err)
-		return nil, nil
-	}
+	runClamavScan(clamavBin, archiveDir, "archive", pass)
```
academo, line 84: we want to keep the logme debug log for future debugging (here and all other passes)

it is also a common convention in go to return an error in a function and check its output

academo (commented):
overall the idea is sound but the code implementation is removing log lines, re-implementing testing patterns and changing some parts of the code that doesn't need to change.

pkg/analysis/analysis.go
tolzhabayev: This is used at runtime by ReportIncomplete to check whether all rules for the current analyzer are disabled https://github.com/grafana/plugin-validator/blob/41824b0/pkg/analysis/incomplete.go#L7-L15

pkg/analysis/passes/coderules/coderules.go
tolzhabayev: Errors contains scan failures, such as parsing failures or OOM. Findings are in Results, which we still report below.

pkg/analysis/passes/gosec/gosec.go
tolzhabayev: should be, see TestQuietSuccessfulScan and TestIncompleteScan -> exit_1

academo (approved):

